PR #20, opened 2024-05-22
Add support for text

### LINKED ISSUE #14: Text

Pante:
Fixes #14

### DIFF
--- a/forui/lib/src/theme/style.dart
+++ b/forui/lib/src/theme/style.dart
@@ -4,27 +4,33 @@ import 'package:flutter/widgets.dart';
 /// The overarching style that is used to configure the properties of widget-specific styles if they are not provided.
 final class FStyle with Diagnosticable {
 
-  /// The border radius.
-  final BorderRadius borderRadius;
-
   /// The text style.
   final TextStyle textStyle;
 
+  /// The border radius.
+  final BorderRadius borderRadius;
+
   /// Creates an [FStyle].
-  FStyle({BorderRadius? borderRadius, TextStyle? textStyle}):
-    borderRadius = borderRadius ?? BorderRadius.circular(8),
-    textStyle = textStyle ?? const TextStyle(fontSize: 10);
+  FStyle({required this.textStyle, BorderRadius? borderRadius}):
+    borderRadius = borderRadius ?? BorderRadius.circular(8);
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
-    properties.add(DiagnosticsProperty<BorderRadius>('borderRadius', borderRadius));
+    properties
+      ..add(DiagnosticsProperty<TextStyle>('textStyle', textStyle))
+      ..add(DiagnosticsProperty<BorderRadius>('borderRadius', borderRadius));
   }
 
   @override
-  bool operator ==(Object other) => identical(this, other) || other is FStyle && borderRadius == other.borderRadius;
+  bool operator ==(Object other) =>
+      identical(this, other) ||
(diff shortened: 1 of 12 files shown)


forui/lib/src/widgets/card/card_style.dart
```diff
@@ -1,21 +1,44 @@
...
 
 /// [FCard]'s style.
-class FCardStyle {
+final class FCardStyle with Diagnosticable {
   /// The decoration.
   final BoxDecoration decoration;
 
+  /// The body.
+  final TextStyle text;
+
```
kawaijoe, line 10: Where is this used?

forui/lib/src/theme/theme.dart
```diff
@@ -3,13 +3,10 @@ import 'package:flutter/material.dart';
...
-/// Represents a ForUI theme.
+/// Represents a Forui theme.
+///
+/// See [ThemeBuildContext.theme] for accessing the current theme.
 class FTheme extends StatelessWidget {
-  /// Retrieves the theme data.
-  static FThemeData of(BuildContext context) {
-    final theme = context.dependOnInheritedWidgetOfExactType<_InheritedTheme>();
-    return theme?.data ?? FThemes.zinc.light;
-  }
```
kawaijoe, line 12: I think this should stay. It's a pretty standard pattern to include the `of` method.

kawaijoe (changes requested):

forui/lib/src/theme/style.dart
```diff
@@ -7,8 +7,13 @@ final class FStyle with Diagnosticable {
...
   final BorderRadius borderRadius;
 
+  /// The text style.
+  final TextStyle textStyle;
+
   /// Creates an [FStyle].
-  FStyle({BorderRadius? borderRadius}) : borderRadius = borderRadius ?? BorderRadius.circular(8);
+  FStyle({BorderRadius? borderRadius, TextStyle? textStyle}):
+    borderRadius = borderRadius ?? BorderRadius.circular(8),
+    textStyle = textStyle ?? const TextStyle(fontSize: 10);
```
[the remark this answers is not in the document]
Pante: I personally think it should be stored in `FStyle`. Font should be responsible for only font-related information (not text styling). IMO this is consistent with how the other styles store textstyles if we think of `FStyle` as being the overarching style for the application.

forui/lib/src/widgets/card/card_style.dart
Pante: This is used in https://github.com/forus-labs/forui/blob/9154bd8e85fef5fdd68b40edef3cbd7f89424a48/forui/lib/src/widgets/card/card.dart#L36.

kawaijoe: I see, if that's the case we should remove this. It's largely pointless imo and might result in unexpected TextStyle. I'd assume most users will think a `Text` will behave the same anywhere throughout the application.

forui/lib/src/theme/style.dart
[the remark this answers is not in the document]
kawaijoe: Placement of the TextStyle aside, the font size doesn't really do much alone. We need to set the color and probably a list of TextStyles for different situations. I have strong doubts if TextStyle should even be offered in this manner.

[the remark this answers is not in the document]
kawaijoe: It could probably make more sense if you want to include constant int for the font size, so we can use it similar to how tailwind defines font sizes.

https://tailwindcss.com/docs/font-size

kawaijoe (approved):

PR #21, opened 2024-05-23
Add separator

### LINKED ISSUE #17: Separator

Pante:
Fixes #17 

* Light horizontal

* Dark horizontal

* Light vertical

* Dark vertical

### DIFF
--- a/forui/lib/src/theme/theme_data.dart
+++ b/forui/lib/src/theme/theme_data.dart
@@ -19,6 +19,12 @@ class FThemeData with Diagnosticable {
   /// The card style.
   final FCardStyle cardStyle;
 
+  /// The separator style.
+  final FSeparatorStyle separatorStyle;
+
+  /// The vertical separator style.
+  final FSeparatorStyle verticalSeparatorStyle;
+
   /// Creates a [FThemeData].
   FThemeData({
     required this.colorScheme,
@@ -26,6 +32,8 @@ class FThemeData with Diagnosticable {
     required this.style,
     required this.boxStyle,
     required this.cardStyle,
+    required this.separatorStyle,
+    required this.verticalSeparatorStyle,
   });
 
   /// Creates a [FThemeData] that inherits the given arguments' properties.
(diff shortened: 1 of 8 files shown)


forui/lib/src/widgets/separator/separator.dart
```diff
@@ -0,0 +1,84 @@
...
+
+}
+
+/// [FSeparator]'s style.
+final class FSeparatorStyle with Diagnosticable {
+
+  /// The default padding for horizontal and vertical separators.
+  static const defaultPadding = (
+    horizontal: EdgeInsets.only(top: 20, bottom: 20),
+    vertical: EdgeInsets.only(left: 20, right: 20),
```
kawaijoe, line 47: Nitpick but I think we can use, `EdgeInsets.symmetric`.

forui/lib/src/theme/theme_data.dart
```diff
@@ -19,13 +19,21 @@ class FThemeData with Diagnosticable {
   /// The card style.
   final FCardStyle cardStyle;
 
+  /// The separator style.
+  final FSeparatorStyle separatorStyle;
+
+  /// The vertical separator style.
+  final FSeparatorStyle verticalSeparatorStyle;
```
kawaijoe, line 26: Should we combine this into a single class similar to what we're planning to do with `FButtonStyles`? For consistency reasons.

kawaijoe (changes requested):

forui/lib/src/widgets/separator/separator.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// The padding surrounding the separating line.
+  final EdgeInsetsGeometry padding;
+
+  /// The thickness of the separating line. Defaults to 1.
+  ///
+  /// ## Contract:
+  /// Throws an [AssertionError] if:
+  /// * `thickness` <= 0.0
+  /// * `thickness` is Nan
+  final double thickness;
```
kawaijoe, line 62: Do you think it's a good idea to add this into the global style? (in addition to `borderRadius`) Could be useful for `Card` and `Button` too.

Also, it might be a good idea to follow the CSS naming convention for this field:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-width
https://tailwindcss.com/docs/border-width

kawaijoe (approved):

PR #31, opened 2024-06-01
Add switch

### LINKED ISSUE #29: Switch

Pante:
Closes #29


forui/lib/src/widgets/switch.dart
```diff
@@ -0,0 +1,177 @@
...
+  /// drag behavior feel slightly more reactive.
+  ///
+  /// By default, the drag start behavior is [DragStartBehavior.start].
+  final DragStartBehavior dragStartBehavior;
+
+  /// Creates a [FSwitch].
+  const FSwitch({
+    required this.value,
+    required this.onChanged,
+    super.key,
```
kawaijoe, line 82: Should super be the last element?

forui/lib/src/widgets/switch.dart
```diff
@@ -0,0 +1,177 @@
...
+  final FocusNode? focusNode;
+
+  /// Handler called when the focus changes.
+  ///
+  /// Called with true if this widget's node gains focus, and false if it loses focus.
+  final ValueChanged<bool>? onFocusChange;
+
+  /// Determines the way that drag start behavior is handled.
+  ///
+  /// If set to [DragStartBehavior.start], the drag behavior used to move the
```
kawaijoe, line 66: This feels kind of confusing tbh

forui/lib/src/widgets/switch.dart
```diff
@@ -0,0 +1,177 @@
...
+    required this.thumb,
+    required this.focus,
+  });
+
+  /// Creates a [FSwitchStyle] that inherits its properties from [colorScheme].
+  FSwitchStyle.inherit({required FColorScheme colorScheme})
+      : checked = colorScheme.primary,
+        unchecked = colorScheme.border,
+        thumb = colorScheme.background,
+        focus = HSLColor.fromColor(colorScheme.primary.withOpacity(0.80))
```
kawaijoe, line 149: Do we need to convert this to `HSLColor`?

forui/lib/src/widgets/switch.dart
Pante: It was copied from CupertinoSwitch. Out of curiosity, which part is confusing?

forui/lib/src/widgets/switch.dart
Pante: Good catch!

forui/lib/src/widgets/switch.dart
Pante: This code is also copied from CupertinoSwitch. There probably is a way to do it but I'm not too sure.

PR #33, opened 2024-06-03
Tidy up documentation and expose more fields on button

Pante:


forui/lib/src/foundation/tappable.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
```
kawaijoe, line 4: Do we need both `material.dart` and `cupertino.dart`?

forui/lib/src/widgets/button/button.dart
```diff
@@ -18,34 +18,70 @@ part 'button_content_style.dart';
...
+  /// Called with true if this widget's node gains focus, and false if it loses focus.
+  final ValueChanged<bool>? onFocusChange;
 
   /// The builder.
   final Widget Function(BuildContext, FButtonStyle) builder;
 
-  /// Creates a [FButton] widget.
+  /// Creates a [FButton].
   FButton({
+    required this.onPress,
```
kawaijoe, line 60: I don't think `onPress` should be required.

forui/lib/src/widgets/button/button_content.dart
```diff
@@ -31,14 +20,16 @@ final class FButtonContent extends StatelessWidget {
...
     return Padding(
       padding: style.padding,
       child: Row(
         mainAxisAlignment: MainAxisAlignment.center,
         children: [
           if (icon != null) ...[
-            icon!(height: 20, colorFilter: enabled ? style.enabledIcon : style.disabledIcon),
+            icon!(
+              height: 20,
+              colorFilter: ColorFilter.mode(enabled ? style.enabledIcon : style.disabledIcon, BlendMode.srcIn),
```
kawaijoe, line 31: Why not just store the ColorFilter object?

kawaijoe (changes requested):

forui/lib/src/widgets/button/button.dart
Pante: I think it should be, it's required for buttons and switches in material & cuptertino.

forui/lib/src/widgets/button/button_content.dart
Pante: I thought about it, and it felt more convenient to expose a color rather than a `ColorFilter`. Most users will only be interested in changing the color and not the `BlendMode`.

kawaijoe (approved):

PR #34, opened 2024-06-06
Add TextField

### LINKED ISSUE #10: Input

Pante:
This PR introduces `FTextField` which is based on Material's `FTextField`. It provides a few named constructors for common use-cases. `FTextField` differs slightly from Shadui's `TextArea` as it does not contain a button in the bottom right corner for resizing the `FTextField`.

It also lays the groundwork for changing its visual appearance when an error occurs. This is not automatically configured since we currently lack a global error-related color.

Closes #10, #18.


forui/lib/src/theme/theme_data.dart
```diff
@@ -46,6 +49,7 @@ class FThemeData with Diagnosticable {
     required this.boxStyle,
     required this.separatorStyles,
     required this.switchStyle,
+    required this.textFieldStyle,
```
kawaijoe, line 52: We should order these based on how the folders are ordered. Eg. in this case, `textFieldStyle` should come after `headerStyle`

forui/lib/src/widgets/text_field/text_field_style.dart
```diff
@@ -0,0 +1,234 @@
...
+      font: font,
+      style: style,
+    ),
+    error = FTextFieldStateStyle.inherit( // TODO: add error colors
+      contentColor: colorScheme.primary,
+      hintColor: colorScheme.mutedForeground,
+      borderColor: colorScheme.border,
+      font: font,
+      style: style,
+    ),
```
kawaijoe, line 96: I think we have generic error colors currently in the theme. We could potentially add those in?

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -0,0 +1,691 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
```
kawaijoe, line 4: Do we required both `material.dart` and `cupertino.dart`?

forui/lib/src/widgets/text_field/text_field_style.dart
```diff
@@ -0,0 +1,234 @@
...
+    focused = FTextFieldStateStyle.inherit(
+      contentColor: colorScheme.primary,
+      hintColor: colorScheme.mutedForeground,
+      borderColor: colorScheme.primary,
+      font: font,
+      style: style,
+    ),
+    focusedError = FTextFieldStateStyle.inherit( // TODO: add error colors
```
kawaijoe, line 104: Could probably follow a similar default as above.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -0,0 +1,691 @@
...
+  /// misspelled words.
+  ///
+  /// If the [SpellCheckService] is left null, spell check is disabled by default unless the [DefaultSpellCheckService]
+  /// is supported, in which case it is used. It is currently supported only on Android and iOS.
+  ///
+  /// If this configuration is left null, then spell check is disabled by default.
+  final SpellCheckConfiguration? spellCheckConfiguration;
+
+  /// The label.
+  final String? label;
```
kawaijoe, line 423: Since we're providing a label, does it make sense to include a field for an error message?

forui/lib/src/widgets/text_field/text_field.dart
Pante: Yes we do, we currently default text selection & text selection control colors to `CupertinoColors.activeBlue`.

forui/lib/src/widgets/text_field/text_field.dart
Pante: I was thinking of addressing it in a future PR since we didn't have the error colors sorted out.

kawaijoe (approved):

Issue #40, opened 2024-06-14
Add accessibility features to `FButton`


Pante:
The current `FButton` implementation does not support accessibility tools such as screen readers.

PR #41, opened 2024-06-14
Add FDialog

### LINKED ISSUE #39: Dialogs

Pante:
Closes #39, #40


forui/lib/src/widgets/button/button_content.dart
```diff
@@ -48,3 +48,60 @@ part of 'button.dart';
...
+  FButtonContentStyle({
+    required this.enabledText,
+    required this.disabledText,
+    required this.enabledIcon,
+    required this.disabledIcon,
+    required this.padding,
+  });
+
+  /// Creates a [FButtonContentStyle] that inherits its properties from the given [foreground] and [disabledForeground].
+  FButtonContentStyle.inherit({ required FTypography typography,required Color foreground, required Color disabledForeground})
```
kawaijoe, line 79:
```suggestion
  FButtonContentStyle.inherit({required FTypography typography, required Color foreground, required Color disabledForeground})
```

forui/lib/src/widgets/dialog/dialog.dart
```diff
@@ -0,0 +1,193 @@
...
+            ),
+            child: Semantics(
+              scopesRoute: true,
+              explicitChildNodes: true,
+              namesRoute: true,
+              label: semanticLabel,
+              child: ConstrainedBox(
+                constraints: const BoxConstraints(
+                  minWidth: 280.0,
+                  maxWidth: 560.0,
```
kawaijoe, line 116: Should we place this in styles instead of leaving them as magic values?

forui/lib/src/widgets/dialog/dialog_content.dart
```diff
@@ -0,0 +1,178 @@
+part of 'dialog.dart';
+
+@internal sealed class FDialogContent extends StatelessWidget {
+  final FDialogContentStyle style;
+  final CrossAxisAlignment alignment;
+  final String? title;
+  final TextAlign titleTextAlign;
+  final String? subtitle;
+  final TextAlign subtitleTextAlign;
```
kawaijoe, line 9: Feels a bit weird to offer `alignment` and `titleTextAlign`/`subtitleTextAlign`. Are there cases where we would like to use all three of the fields?

forui/test/golden/badge/zinc-dark-FBadgeVariant.destructive-badge-content.png
kawaijoe, line 1: On a side note, we need to devise a solution to prevent everything from regenerating on every PR.

kawaijoe (changes requested):

forui/lib/src/widgets/dialog/dialog_content.dart
```diff
@@ -0,0 +1,178 @@
...
+  final List<Widget> actions;
+
+  const FDialogContent({
```
kawaijoe, line 12: Should we add a child?

Pante: But where should the child be placed?

forui/lib/src/widgets/dialog/dialog.dart
Pante: I personally don't think these values should be configurable.

forui/lib/src/widgets/dialog/dialog_content.dart
Pante: These aren't exposed directly in the style. We expose these 3 fields since the alignment of the title/subtitle is towards the left in a horizontal layout, and centered in a vertical layout.

forui/lib/src/widgets/dialog/dialog_content.dart
kawaijoe: Between the button and subtitle

forui/lib/src/widgets/dialog/dialog.dart
kawaijoe: Would it be possible if a content requires more than the max width?

kawaijoe (approved):

PR #49, opened 2024-06-22
Typography revamp

kawaijoe:


forui/lib/src/theme/theme.dart
```diff
@@ -99,25 +100,25 @@ class FTheme extends StatelessWidget {
...
-          fontSize: data.typography.base,
-          color: data.colorScheme.foreground,
+        data: data,
+        child: Directionality(
+          textDirection: textDirection ?? Directionality.of(context),
+          child: DefaultTextStyle(
+            style: data.typography.base.copyWith(
+              fontFamily: data.typography.defaultFontFamily,
+              color: data.colorScheme.foreground,
+            ),
```
Pante, line 110: Should we set the font directly in `base` & other text styles?

Pante (changes requested):
Generally looks good, I think the section on scaling in CONTRIBUTING.md needs to be removed.

Pante (approved):

PR #50, opened 2024-06-24
Core documentation

kawaijoe:
As it might be challenging to read the markdown, I've provided some images for your reference.

# Getting Started

screencapture-localhost-3000-docs-2024-06-24-10_52_21

# Themes
screencapture-localhost-3000-docs-themes-2024-06-24-10_54_20

# Icon
screencapture-localhost-3000-docs-icon-2024-06-24-10_54_40

### DIFF
--- a/.idea/runConfigurations/docs___run.xml
+++ b/.idea/runConfigurations/docs___run.xml
@@ -1,5 +1,9 @@
 <component name="ProjectRunConfigurationManager">
+<<<<<<<< HEAD:.idea/runConfigurations/docs___run.xml
   <configuration default="false" name="docs - run" type="js.build_tools.npm">
+========
+  <configuration default="false" name="Docs - dev" type="js.build_tools.npm">
+>>>>>>>> cce3bd1 (Complete getting started):.idea/runConfigurations/Docs___dev.xml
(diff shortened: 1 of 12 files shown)


Pante (commented):
# Getting Started

## Installation
Should we export `forui_assets` in forui? That way people don't have to install two packages.

## Usage
IMO the first example and the Material/Cupertino section should be merged since most people would need routing.

# Themes

## Usage

In the example, I think we can include the type on the LHS of the assignment rather than the comment.

`final FTheme theme = context.theme;` instead of `final theme = context.theme; // FTheme`.

# Icon

## Usage
I feel like the type on the LHS should be included to make it easier to reference.

kawaijoe:
> Should we export `forui_assets` in forui? That way people don't have to install two packages.

Yep we can export `forui_assets`. There's a callout stating that they don't need to install both if they installed `forui`.

> IMO the first example and the Material/Cupertino section should be merged since most people would need routing.

I disagree, routing can be handled directly with a `Navigator` without `MaterialApp`.
https://api.flutter.dev/flutter/widgets/Navigator-class.html



> `final FTheme theme = context.theme;` instead of `final theme = context.theme; // FTheme`.

Fixed!

PR #73, opened 2024-07-03
Fix recommended usage of FTheme

### LINKED ISSUE #70: Improve routing ergonomics
**Describe the bug**
Since MaterialApp provides routing capabilities and wraps FTheme, new routes do not inherit previously defined theme data.
However, since MaterialApp handles localization and text directionality which FTheme relies on, simply wrapping MaterialApp doesn't seem like an effective solution

**Expected behavior**
There should be a way to provide FTheme to all routes under MaterialApp.

**Screenshots**
image

Consequently, Dialogs doe not inherit the right FTheme data.

Pante:
Fixes #70

### DIFF
--- a/docs/pages/docs/index.mdx
+++ b/docs/pages/docs/index.mdx
@@ -73,9 +73,10 @@ This is especially useful when you need to use a Material/Cupertino widget that 
 @override
 Widget build(BuildContext context) => MaterialApp(
   // Other configurations...
-  home: FTheme(
+  builder: (context, child) => FTheme(
     data: FThemes.zinc.light,
-    child: const FScaffold(...),
+    child: c
   ),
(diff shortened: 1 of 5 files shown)


docs/pages/docs/index.mdx
```diff
@@ -73,9 +73,10 @@ This is especially useful when you need to use a Material/Cupertino widget that
 @override
 Widget build(BuildContext context) => MaterialApp(
   // Other configurations...
-  home: FTheme(
+  builder: (context, child) => FTheme(
     data: FThemes.zinc.light,
-    child: const FScaffold(...),
+    child: c
```
kawaijoe, line 78: Is this supposed to be `child`?

PR #74, opened 2024-07-03
Integrate switches with forms

Pante:
This PR re-implements switches to be usable in forms. It also polishes checkbox's rough edges.

### DIFF
--- a/docs/pages/docs/text-field.mdx
+++ b/docs/pages/docs/text-field.mdx
@@ -135,7 +135,7 @@ FTextField.multiline(
 
 <Tabs items={['Preview', 'Code']}>
   <Tabs.Tab>
-    <Widget name='text-field' variant='form' query={{}}/>
+    <Widget name='text-field' variant='form' height={550} query={{}} height={550}/>
   </Tabs.Tab>
   <Tabs.Tab>
     ```dart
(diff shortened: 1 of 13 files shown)


docs/pages/docs/text-field.mdx
```diff
@@ -135,7 +135,7 @@ FTextField.multiline(
 
 <Tabs items={['Preview', 'Code']}>
   <Tabs.Tab>
-    <Widget name='text-field' variant='form' query={{}}/>
+    <Widget name='text-field' variant='form' height={550} query={{}} height={550}/>
```
kawaijoe, line 138: I don't think we need to specify height twice.

forui/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Next
+
+### Enhancements
+* **Breaking** Change `FSwitch` to be usable in `Form`s.
```
kawaijoe, line 4: Is this actually a breaking change?

samples/lib/widgets/switch.dart
```diff
@@ -7,22 +7,81 @@ import 'package:forui_samples/sample_scaffold.dart';
...
       padding: const EdgeInsets.all(16),
-      child: ValueListenableBuilder(
-        valueListenable: notifier,
-        builder: (context, value, __) => FSwitch(
-          value: value,
-          onChanged: (value) => notifier.value = value,
-        ),
+      child: FSwitch(
+        enabled: enabled,
       ),
```
kawaijoe, line 22: Nitpick, but I think the parameter can be on one line.

kawaijoe (changes requested):

Issue #58, opened 2024-06-28
Add padding constants


kawaijoe:

kawaijoe:
Accidentally created a duplicate: #91.

Will be closing this issue.

PR #90, opened 2024-07-17
Add padding constants #58

JoDeveloper:


kawaijoe:
Hey @JoDeveloper!

Thank you for your contribution. I've decided to close this pr till we can iron out the direction we would like to take with constant values.

I have described our dilemma and considerations in detail at #91. Would love to hear your thoughts on it!

PR #132, opened 2024-08-01
Fix `FThemeData.copyWith(...)`

### LINKED ISSUE #130: Theme customization doesn't work
**Describe the bug**
I tried to change the crown parameters using methods as in the [documentation via copyWith](https://forui.dev/docs/themes#customize-themes). The problem is that the values ​​in the FThemeData object change, but they don't work in the application itself.

**To Reproduce**
1. Create a flutter project
2. Add the following code

**Expected behavior**
The colors should change, the borderRadius should become 0, the font should change.

**Screenshots**
In my implementation nothing was applied, the color remained the same, the corners remained rounded.

image

**Device:**
 - Flutter version: 3.22.2 
 - Forui version: 0.3.0
 - Device: Emulator Pixel 8
 - OS: Android 15 (API 35)

**Additional context**

main.dart
```
void main() {
  WidgetsFlutterBinding.ensureInitialized();

  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  final _router = AppRouter();

  @override
  Widget build(BuildContext context) {
    return MaterialApp.router(
      title: 'MyApp',
      routerConfig: _router.config(),
      builder: (context, child) => FTheme(
        data: FThemes.zinc.dark.copyWith(
          colorScheme: FThemes.zinc.dark.colorScheme.copyWith(
            primary: const Color(0xFFFF2222),
            primaryForeground: const Color(0xFFFFFFFF),
          ),
          typography: FThemes.zinc.dark.typography.copyWith(
            defaultFontFamily: 'GolosText',
          ),
          style: FThemes.zinc.dark.style.copyWith(
            borderRadius: BorderRadius.zero,
          ),
        ),
        child: child!,
      ),
    );
  }
}
```

button implementation
```
class AccountPage extends StatelessWidget {
  const AccountPage({super.key});

  @override
  Widget build(BuildContext context) {
    return FScaffold(
        content: Center(
            child: FButton(
              label: const Text('Button'),
              onPress: () {},
            )
        )
    );
  }
}
```

Pante:
Fixes #130

### DIFF
--- a/docs/pages/docs/themes.mdx
+++ b/docs/pages/docs/themes.mdx
@@ -20,7 +20,7 @@ Our theming solution is designed to help you get started quickly while offering 
 
 <Callout type="info">
     Forui does not manage the theme brightness (light or dark) automatically.
-    You need to specify the theme explicitly in `FTheme()`.
+    You need to specify the theme explicitly in `FTheme(...)`.
 
     ```dart filename="main.dart" {3}
     @override
@@ -75,10 +75,10 @@ A more detailed explanation of each class can be found in the [Class Diagram](#c
 
 ## Customize Themes
 
-It's recommended to use the predefined themes as a starting point and customize them with the `inhert(...)` constructor
-and copyWith(...)` method to suit your needs.
+It's recommended to use the predefined themes as a starting point and customize them with the `inhert(...)`[] constructor
+and `copyWith(...)` method to suit your needs.
 
-```dart filename="main.dart" {3-4, 8, 11, 14}
+```dart filename="main.dart" {3-14, 17-23}
 @override
 Widget build(BuildContext context) {
   final theme = FThemeData.inherit(
--- a/docs/pages/docs/themes.mdx
+++ b/docs/pages/docs/themes.mdx
@@ -75,30 +75,36 @@ A more detailed explanation of each class can be found in the [Class Diagram](#c
 
 ## Customize Themes
 
-It's recommended to use the predefined themes as a starting point and customize them with the `copyWith(...)` method to suit your needs.
+It's recommended to use the predefined themes as a starting point and customize them with the `inhert(...)` constructor
+and copyWith(...)` method to suit your needs.
 
 ```dart filename="main.dart" {3-4, 8, 11, 14}
 @override
-Widget build(BuildContext context) => FTheme(
-      data: FThemes.zinc.light.copyWith(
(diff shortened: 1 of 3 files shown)


docs/pages/docs/themes.mdx
```diff
@@ -75,30 +75,36 @@ A more detailed explanation of each class can be found in the [Class Diagram](#c
...
+      defaultFontFamily: 'Roboto',
+    ).scale(sizeScalar: 0.8),
+    style: FThemes.zinc.light.style.copyWith(
+      borderRadius: BorderRadius.zero,
+    ),
+  );
+
+  return FTheme(
+    data: theme.copyWith(
+      cardStyle: theme.cardStyle.copyWith(
```
kawaijoe, line 99: Might be a good idea to add a call out above, that the copyWith here is with the `theme` and not `FThemes.zinc`. A quick explanation help!

docs/pages/docs/themes.mdx
```diff
@@ -75,30 +75,36 @@ A more detailed explanation of each class can be found in the [Class Diagram](#c
...
+      primary: const Color(0xFF0D47A1), // dark blue
+      primaryForeground: const Color(0xFFFFFFFF), // white
+    ),
+    typography: FThemes.zinc.light.typography.copyWith(
+      defaultFontFamily: 'Roboto',
+    ).scale(sizeScalar: 0.8),
+    style: FThemes.zinc.light.style.copyWith(
+      borderRadius: BorderRadius.zero,
+    ),
+  );
```
kawaijoe, line 95: Is it better for this to be outside the build method? Possibly as a constant.

Pante: I just figured it'll be easier to put everything inside the build method for the example.

docs/pages/docs/themes.mdx
Pante: Good catch!

docs/pages/docs/themes.mdx
kawaijoe: I don't think it would be too long if we include a stateless widget. I think most people would copy from the docs directly and it would be better if it isn't placed in the build widget. No strong opinions on it though.

Pante: Mmmm fair enough

PR #176, opened 2024-08-30
Add tooltip

Pante:
**Describe the changes**

**Checklist**


forui/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,391 @@
...
+          onExit: (_) => _exit(),
+          // We have to use a Listener as GestureDetector's arena implementation allows only 1 gesture to win. It is
+          // problematic if the child is a button. See https://github.com/flutter/flutter/issues/92103.
+          child: Listener(
+            onPointerDown: (_) => _exit(),
+            child: child,
+          ),
+        );
+      }
+    }
```
kawaijoe, line 241: Is there a point checking for platform? It could probably make more sense for long tap + hover to activate the tooltip. That way the usage is consistent and we don't have to deal with awkward use cases where eg. touchscreen windows laptop.

forui/test/src/widgets/resizable/resizable_test.dart
```diff
@@ -25,6 +26,8 @@ void main() {
   }
 
   testWidgets('vertical drag downwards', (tester) async {
+    debugDefaultTargetPlatformOverride = TargetPlatform.linux;
```
kawaijoe, line 29: Just curious, why do we want to target this for linux?

Pante: No reason really, we can target any desktop platform.

forui/lib/src/widgets/tooltip.dart
Pante: I don’t have any strong opinion on it, I just thought that it might be weird if you could long press to view the tooltip on desktop.

forui/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,396 @@
...
+enum FToolTipBehavior {
+  /// Denotes that the tooltip is shown when hovered over.
+  hover,
+
+  /// Denotes that the tooltip is shown when hovered over or long pressed.
+  hoverOrLongPress,
+
+  /// Denotes that the tooltip is shown when long pressed.
+  longPress,
+}
```
kawaijoe, line 75: Is using 2 boolean cleaner?

forui/lib/src/widgets/tooltip.dart
```diff
@@ -0,0 +1,396 @@
...
+      blurRadius: 6,
+      spreadRadius: -1,
+    ),
+    BoxShadow(
+      color: Color(0x1a000000),
+      offset: Offset(0, 2),
+      blurRadius: 4,
+      spreadRadius: -2,
+    ),
+  ];
```
kawaijoe, line 322: Is this intended? They wouldn't be able to change the values of static variables.

forui/test/src/widgets/resizable/resizable_test.dart
kawaijoe: Rather, why do we need to target desktop platforms?

forui/test/src/widgets/tooltip/tooltip_test.dart
```diff
@@ -0,0 +1,287 @@
...
+        await gesture.moveTo(tester.getCenter(find.byType(FButton)));
+        await tester.pumpAndSettle(const Duration(seconds: 1));
+
+        expect(find.text('tip'), findsNothing);
+
+        debugDefaultTargetPlatformOverride = null;
+      });
+
+      testWidgets('hover shows tooltip', (tester) async {
+        debugDefaultTargetPlatformOverride = TargetPlatform.linux;
```
kawaijoe, line 135: Same here, does the platform matter? Though it could be interesting to see if this could be used to avoid the slight variances between how text is rendered between platforms.

forui/test/src/widgets/resizable/resizable_test.dart
Pante: This was done previously to test the platform specific stuff, but I guess it isn’t needed anymore

forui/test/src/widgets/tooltip/tooltip_test.dart
Pante: The platform matters for Resizable since there's a difference in hit area for the different platforms.

forui/lib/src/widgets/tooltip.dart
Pante: Yup it's intended. It's a constant value that they can use in their box decoration.

forui/lib/src/widgets/tooltip.dart
Pante: I personally think that using 2 booleans is not a good idea since it's possible to pass in `false` for both while also not passing it a controller to manually show the tooltip.

kawaijoe: Actually isn't that a great idea, we don't need the manual constructor that way.

kawaijoe (approved):

PR #188, opened 2024-09-11
Accordion

Daviiddoo:
**Describe the changes**

**Checklist**


forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+  double get value => _expand.value;
+
+  @override
+  void dispose() {
+    _animation.dispose();
+    super.dispose();
+  }
+}
+
+/// A vertically stacked set of interactive headings that each reveal a section of content.
```
Pante, line 74:
```suggestion
/// An interactive heading that reveals a section of content.
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+import 'package:flutter_svg/flutter_svg.dart';
+import 'dart:math' as math;
+
+import 'package:forui/forui.dart';
+import 'package:forui/src/foundation/tappable.dart';
+import 'package:forui/src/foundation/util.dart';
+import 'package:meta/meta.dart';
+
+/// A controller that stores the expanded state of an [FAccordion].
+class FAccordionController extends ChangeNotifier {
```
Pante, line 14: For consistency, I think "expand" should be replaced with "show" while "close" should be replaced with "hide".

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+    );
+    _expand = Tween<double>(
+      begin: 0,
+      end: 100,
+    ).animate(
+      CurvedAnimation(
+        curve: Curves.ease,
+        parent: _animation,
+      ),
+    )..addListener(notifyListeners);
```
Pante, line 37: Why do we need to add a listener here?

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+  Future<void> close() async {
+    await _animation.reverse();
+    notifyListeners();
+  }
+
+  /// True if the accordion is expanded. False if it is closed.
+  bool get expanded => _animation.value == 1.0;
+
+  /// The animation value.
+  double get value => _expand.value;
```
Pante, line 65: Do we use this value for anything? If not, I don't think it needs to be exposed.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+  final bool initiallyExpanded;
+
+  /// The child.
+  final Widget child;
+
+  /// Creates an [FAccordion].
+  const FAccordion({
+    required this.child,
+    this.style,
+    this.title = '',
```
Pante, line 98: Should this be a widget instead?

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+    this.controller,
+    this.initiallyExpanded = true,
+    super.key,
+  });
+
+  @override
+  //ignore:library_private_types_in_public_api
+  _FAccordionState createState() => _FAccordionState();
```
Pante, line 106: This should be State<FAccordion>, and the ignore comment removed.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties
+      ..add(DiagnosticsProperty('style', style))
+      ..add(StringProperty('title', title))
+      ..add(DiagnosticsProperty('controller', controller))
+      ..add(DiagnosticsProperty('initiallyExpanded', initiallyExpanded));
```
Pante, line 115: use a flag property instead.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+  final FAccordionStyle? style;
+
+  /// The title.
+  final String title;
+
+  /// The accordion's controller.
+  final FAccordionController? controller;
+
+  /// Whether the accordion is initially expanded. Defaults to true.
+  final bool initiallyExpanded;
```
Pante, line 89: Why is initiallyExpanded declared on both the widget and controller?

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+      _controller.dispose();
+    }
+    super.dispose();
+  }
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DiagnosticsProperty('controller', _controller));
+  }
```
Pante, line 217:
```suggestion
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,399 @@
...
+  final double percentage;
+
+  _Clipper({required this.percentage});
+
+  @override
+  Rect getClip(Size size) => Offset.zero & Size(size.width, size.height * percentage);
+
+  @override
+  bool shouldReclip(covariant _Clipper oldClipper) => oldClipper.percentage != percentage;
+}
```
Pante, line 399: Can this be moved above the style?

forui/lib/widgets/accordion.dart
```diff
@@ -0,0 +1,8 @@
+/// {@category Widgets}
+///
+/// An image element with a fallback for representing the user.
+///
+/// See https://forui.dev/docs/avatar for working examples.
```
Pante, line 5: You forgot to update the comment.

Pante (changes requested):

forui/lib/src/widgets/accordion.dart
Daviiddoo: we use this to supply the animation value to the RenderObject and Clipper

Pante: Ah fair enough.

forui/lib/src/widgets/accordion.dart
Daviiddoo: I added this to allow the user to set the `initiallyExpanded` field of the default controller,  only if they did not add in a their own controller. 

After your question, I was wondering if adding some new comments will help to make that clearer.
  /// Whether the accordion is initially expanded. Defaults to true.
  ///
  /// This flag will be ignored if a [controller] is provided.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,395 @@
...
+    super.dispose();
+  }
+}
+
+/// An interactive heading that reveals a section of content.
+///
+/// See:
+/// * https://forui.dev/docs/accordion for working examples.
+class FAccordion extends StatefulWidget {
+  /// The accordion's style. Defaults to the appropriate style in [FThemeData.accordionStyle].
```
Pante, line 81:
```suggestion
  /// The accordion's style. Defaults to [FThemeData.accordionStyle].
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,395 @@
...
+  final FAccordionStyle? style;
```
Pante, line 82: nitpick: i think it'll be neater if the fields are arranged in the following order:
* controller
* style
* other fields

forui/lib/src/widgets/accordion.dart
Daviiddoo: When I remove it, the accordion seems to jump between the states abruptly without any animation.

forui/lib/src/widgets/accordion.dart
Pante: I'm kinda on the fence about this. On one hand, I agree it's much more convenient to configure it via a bool param. On the other, specifying it in both the controller and widget seems confusing.

Perhaps we can make the field nullable, add an assertion in the constructor to ensure that only either param is set?

Daviiddoo: Also, Maybe i should change the controller field to be required? since the widget already provides a default value?

docs/pages/docs/accordion.mdx
```diff
@@ -0,0 +1,103 @@
...
+      title: const Text('Is it accessible?'),
+      child: const Text('Yes. It adheres to the WAI-ARIA design pattern.'),
+    ),
+  ],
+),
+```
+
+## Examples
+
+### With FRadioAccordionController
```
Pante, line 67:
```suggestion
### With Radio Behaviour
```

forui/lib/src/theme/theme_data.dart
```diff
@@ -25,6 +25,9 @@ final class FThemeData with Diagnosticable {
   /// The style. It is used to configure the miscellaneous properties, such as border radii, of Forui widgets.
   final FStyle style;
 
+  /// The accordion styles.
```
Pante, line 28:
```suggestion
  /// The accordion style.
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+import 'package:flutter/widgets.dart';
+
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:meta/meta.dart';
+
+import 'package:forui/forui.dart';
+import 'package:forui/src/foundation/tappable.dart';
+import 'package:forui/src/foundation/util.dart';
+
+/// A controller that stores the expanded state of an [_Item].
```
Pante, line 14:
```suggestion
/// A controller that controls which sections are shown and hidden.
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+interface class FAccordionController extends ChangeNotifier {
+  /// The duration of the expansion and collapse animations.
+  final Duration duration;
```
Pante, line 17: nitpick: this should be renamed to `animationDuration`.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  })  : _min = min,
+        _max = max,
+        _controllers = [],
+        _values = values ?? {},
+        assert(min >= 0, 'The min value must be greater than or equal to 0.'),
+        assert(max == null || max >= 0, 'The max value must be greater than or equal to 0.'),
+        assert(max == null || min <= max, 'The max value must be greater than or equal to the min value.');
+
+  /// Adds an item to the accordion.
+  void addItem(int index, AnimationController controller, Animation expand) {
```
Pante, line 45:
```suggestion
  void addItem(int index, AnimationController controller, Animation animation) {
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+
+import 'package:forui/forui.dart';
+import 'package:forui/src/foundation/tappable.dart';
+import 'package:forui/src/foundation/util.dart';
+
+/// A controller that stores the expanded state of an [_Item].
+interface class FAccordionController extends ChangeNotifier {
+  /// The duration of the expansion and collapse animations.
+  final Duration duration;
+  final List<(AnimationController, Animation)> _controllers;
```
Pante, line 18:
```suggestion
  final List<({AnimationController controller, Animation animation})> _controllers;
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  /// The [min] and [max] values are the minimum and maximum number of selections allowed. Defaults to no minimum or maximum.
+  ///
+  /// # Contract:
+  /// * Throws [AssertionError] if [min] < 0.
+  /// * Throws [AssertionError] if [max] < 0.
+  /// * Throws [AssertionError] if [min] > [max].
+  FAccordionController({
+    int min = 0,
+    int? max,
+    Set<int>? values,
```
Pante, line 34:
```suggestion
    Set<int> values = {},
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+        assert(min >= 0, 'The min value must be greater than or equal to 0.'),
+        assert(max == null || max >= 0, 'The max value must be greater than or equal to 0.'),
+        assert(max == null || min <= max, 'The max value must be greater than or equal to the min value.');
+
+  /// Adds an item to the accordion.
+  void addItem(int index, AnimationController controller, Animation expand) {
+    _controllers.add((controller, expand));
+    if (controller.value == 1.0) {
+      _values.add(index);
+    }
```
Pante, line 49: This logic seems incorrect. What if you:
* Add an item at an index that already exists.
* Add an item at an index that isn't at the end of the list?

Pante: Why do we only add an index to `_values` if the animation == 1, while we unconditionally add both to the list?

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  void removeItem(int index) {
+    _controllers.removeAt(index);
+    _values.remove(index);
+  }
+
+  /// Convenience method for toggling the current expanded status.
+  ///
+  /// This method should typically not be called while the widget tree is being rebuilt.
+  Future<void> toggle(int index) async {
+    if (_controllers[index].$2.value == 100) {
```
Pante, line 62: Should it be 100? I thought it was normally 1.0 for animations.

Pante, line 62: I feel like this method can be tided up.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:meta/meta.dart';
+
+import 'package:forui/forui.dart';
+import 'package:forui/src/foundation/tappable.dart';
+import 'package:forui/src/foundation/util.dart';
+
+/// A controller that stores the expanded state of an [_Item].
+interface class FAccordionController extends ChangeNotifier {
```
Pante, line 15: In hindsight, I think we should make this a `base` class instead since an `interface` class would require them to re-implement `ChangeNotifier` too...

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  /// Hides the content in the accordion.
+  ///
+  /// This method should typically not be called while the widget tree is being rebuilt.
+  Future<void> collapse(int index) async {
+    final controller = _controllers[index].$1;
+    await controller.reverse();
+    notifyListeners();
+  }
```
Pante, line 94: These methods don't perform any min/max checks.

Pante: The document should also be updated to warn that the given index needs to be valid, else an exception will be thrown. Ideally it'll be a `## Contract` section.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  Future<void> toggle(int index) async {
+    final toggle = <Future>[];
+    if (!(_values.length <= _min)) {
+      if (!_values.contains(index) && _values.length >= _max!) {
+        toggle.add(collapse(_values.first));
+        _values.remove(_values.first);
+      }
+    }
+    toggle.add(super.toggle(index));
+    await Future.wait(toggle);
```
Pante, line 117: Please refactor this. It's really hard to follow.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+    await controller.reverse();
+    notifyListeners();
+  }
+}
+
+/// An [FAccordionController] that allows one section to be expanded at a time.
+class FRadioAccordionController extends FAccordionController {
+  /// Creates a [FRadioAccordionController].
+  FRadioAccordionController({
+    int? value,
```
Pante, line 101: We should document what value represents. Perhaps we should change it to initialIndex instead?

forui/lib/src/widgets/accordion.dart
Pante: Should we rename it to `initialIndexes`?

Pante: What if we provide `initialIndexes` that is outside the range of min and max?

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+        toggle.add(collapse(_values.first));
+        _values.remove(_values.first);
+      }
+    }
+    toggle.add(super.toggle(index));
+    await Future.wait(toggle);
+  }
+}
+
+/// A vertically stacked set of interactive headings that each reveal a section of content.
```
Pante, line 121: We're missing a list of links to important/relevant stuff, i.e. docs, controllers, style.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+
+  ///
+  FAccordionItem({required this.title, required this.child, this.initiallyExpanded = false});
+}
+
+/// An interactive heading that reveals a section of content.
+///
+/// See:
+/// * https://forui.dev/docs/accordion for working examples.
+class _Item extends StatefulWidget {
```
Pante, line 202: Thinking about it, should we expose `_Item` instead, and pass in the index via an inherited widget? That way, we don't need the current FAccordionItem.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  /// The title.
+  final Widget title;
+
+  /// The child.
+  final Widget child;
+
+  /// Whether the item is initially expanded.
+  final bool initiallyExpanded;
+
+  ///
```
Pante, line 194: Missing constructor documentation.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+/// An item that represents a header in a [FAccordion].
+class FAccordionItem {
+  /// The title.
+  final Widget title;
+
+  /// The child.
+  final Widget child;
+
+  /// Whether the item is initially expanded.
+  final bool initiallyExpanded;
```
Pante, line 192: Thinking about it, why we do allow people to pass in a list of initialIndexes in the controller when we have this param here? I think the initialIndexes parameter on the controller should be removed.

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+    super.initState();
+    _controller = widget.controller ?? FAccordionController();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final style = widget.style ?? context.theme.accordionStyle;
+    return Column(
+      children: [
+        for (var i = 0; i < widget.items.length; i++)
```
Pante, line 171: nitpick: this can be replaced by 
```dart
for (final (index, widget) in widget.items.indexed)
```

forui/lib/src/widgets/accordion.dart
```diff
@@ -0,0 +1,525 @@
...
+  _Clipper({required this.percentage});
+
+  @override
+  Rect getClip(Size size) => Offset.zero & Size(size.width, size.height * percentage);
+
+  @override
+  bool shouldReclip(covariant _Clipper oldClipper) => oldClipper.percentage != percentage;
+}
+
+/// The [_Item] styles.
```
Pante, line 428: Documentation needs to be updated.

forui/lib/widgets/accordion.dart
```diff
@@ -0,0 +1,8 @@
+/// {@category Widgets}
+///
+/// An interactive heading that reveals a section of content.
```
Pante, line 3: Needs to be updated.

Pante (changes requested):
I haven't reviewed the tests yet since the accordion requires significant rework. Furthermore, I think the accordion file needs to be broken up into multiple files:
* accordion + style
*  item + renderbox
* controllers

forui/lib/src/widgets/accordion.dart
Daviiddoo: The reason I made it nullable is because  `Set<int> values = {}` requires the set to be prefixed with `const` but this makes it unmodifiable later

forui/lib/src/widgets/accordion.dart
Daviiddoo: This can be either 100 when we track the `animation.value`, or 1.0 if we are tracking the `controller.value`.

Which do you prefer?

forui/lib/src/widgets/accordion.dart
Daviiddoo: If so, should `FRadioAccordionController` be a sealed class?

Pante (changes requested):
You're missing unit tests for the controller.

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+            FAccordionItemData(
+              index: index,
+              controller: _controller,
+              child: child,
+            ),
+        ],
+      );
+}
+
+/// The [FAccordion] style.
```
Pante, line 91:
```suggestion
/// The [FAccordion]'s style.
```

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+final class FAccordionStyle with Diagnosticable {
+  /// The title's text style.
```
Pante, line 93:
```suggestion
  /// The title's default text style.
```

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+  final TextStyle childTextStyle;
+
+  /// The padding around the title.
+  final EdgeInsets titlePadding;
+
+  /// The padding around the content.
+  final EdgeInsets childPadding;
+
+  /// The icon.
+  final SvgPicture icon;
```
Pante, line 106: @kawaijoe should we use `SvgPicture` as the type?

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+      icon.hashCode ^
+      dividerColor.hashCode;
+}
+
+@internal
+class FAccordionItemData extends InheritedWidget {
+  @useResult
+  static FAccordionItemData of(BuildContext context) {
+    final data = context.dependOnInheritedWidgetOfExactType<FAccordionItemData>();
+    assert(data != null, 'No FAccordionData found in context');
```
Pante, line 195: ### 
```suggestion
    assert(data != null, 'No FAccordionItemData found in context');
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
+import 'package:flutter/widgets.dart';
+
+/// A controller that controls which sections are shown and hidden.
+class FAccordionController extends ChangeNotifier {
+  /// The duration of the expansion and collapse animations.
```
Pante, line 5:
```suggestion
  /// The duration of the expanding and collapsing animations.
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  final Duration animationDuration;
+
+  /// A list of controllers for each of the headers in the accordion.
```
Pante, line 8:
```suggestion
  /// The animation controllers for each of the sections in the accordion.
  @internal
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  final int _min;
+  final int? _max;
+
+  /// An [FAccordionController] that allows only one section to be expanded at a time.
+  factory FAccordionController.radio({Duration? animationDuration}) => FAccordionController(
+        max: 1,
+        animationDuration: animationDuration ?? const Duration(milliseconds: 200),
+      );
+
+  /// Creates a base [FAccordionController].
```
Pante, line 20:
```suggestion
  /// Creates a [FAccordionController].
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  /// The duration of the expansion and collapse animations.
+  final Duration animationDuration;
+
+  /// A list of controllers for each of the headers in the accordion.
+  final Map<int, ({AnimationController controller, Animation animation})> controllers;
+  final Set<int> _expanded;
+  final int _min;
+  final int? _max;
+
+  /// An [FAccordionController] that allows only one section to be expanded at a time.
```
Pante, line 14: Please document the constructors properly.

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /// Removes an item from the accordion. Returns true if the item was removed.
+  bool removeItem(int index) {
+    final removed = controllers.remove(index);
+    _expanded.remove(index);
+    return removed != null;
+  }
+
+  /// Convenience method for toggling the current expanded status.
```
Pante, line 63:
```suggestion
  /// Convenience method for toggling the current expansion status.
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+
+    if (initiallyExpanded) {
+      if (_max != null && _expanded.length >= _max) {
+        return;
+      }
+      _expanded.add(index);
+    }
+  }
+
+  /// Removes an item from the accordion. Returns true if the item was removed.
```
Pante, line 56:
```suggestion
  /// Removes the item at the given [index] from the accordion. Returns true if the item was removed.
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+    _expanded.remove(index);
+    return removed != null;
+  }
+
+  /// Convenience method for toggling the current expanded status.
+  ///
+  /// This method should typically not be called while the widget tree is being rebuilt.
+  Future<void> toggle(int index) async {
+    final value = controllers[index]?.animation.value;
+
```
Pante, line 68:
```suggestion
    final value = controllers[index]?.animation.value;
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+    if (value == null) {
+      return;
+    }
+
+    value == 100 ? await collapse(index) : await expand(index);
+  }
+
+  /// Shows the content in the accordion.
```
Pante, line 76:
```suggestion
  /// Expands the item at the given [index].
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  ///
+  /// This method should typically not be called while the widget tree is being rebuilt.
+  Future<void> expand(int index) async {
+    if (_expanded.contains(index)) {
+      return;
+    }
+    final futures = <Future<void>>[];
```
Pante, line 83:
```suggestion
    }
    
    final futures = <Future<void>>[];
```

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// An [FAccordionController] that allows only one section to be expanded at a time.
+  factory FAccordionController.radio({Duration? animationDuration}) => FAccordionController(
+        max: 1,
+        animationDuration: animationDuration ?? const Duration(milliseconds: 200),
+      );
+
+  /// Creates a base [FAccordionController].
+  ///
+  /// The [min] and [max] values are the minimum and maximum number of selections allowed. Defaults to no minimum and maximum.
```
Pante, line 22: Please mention that they are inclusive.

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Shows the content in the accordion.
+  ///
+  /// This method should typically not be called while the widget tree is being rebuilt.
+  Future<void> expand(int index) async {
+    if (_expanded.contains(index)) {
+      return;
+    }
+    final futures = <Future<void>>[];
+    if (_expanded.length > _min && _max != null && _expanded.length >= _max) {
```
Pante, line 84:
```suggestion
    if (_max != null && _expanded.length >= _max) {
```

This should be able to be simplified as if _expanded.length is greater than max, it should be greater than min.

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+  Future<bool> _collapse(int index) async {
+    if (_expanded.length <= _min || !_expanded.contains(index)) {
+      return false;
+    }
+
+    _expanded.remove(index);
+
+    await controllers[index]?.controller.reverse();
+    return true;
+  }
```
Pante, line 108: This should be below collapse.

forui/lib/src/widgets/accordion/accordion_controller.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Hides the content in the accordion.
```
Pante, line 110:
```suggestion
  /// Collapses the item at the given [index].
```

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+
+import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
+
+import 'package:forui/forui.dart';
+import 'package:forui/src/foundation/tappable.dart';
+import 'package:forui/src/foundation/util.dart';
+
+@internal
+class FAccordionItemData extends InheritedWidget {
```
Pante, line 11: There shouldn't be two accordion item datas.

Pante: The implementation in the other file should be deleted.

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+
+  @override
+  State<FAccordionItem> createState() => _FAccordionItemState();
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties
+      ..add(DiagnosticsProperty('style', style))
+      ..add(DiagnosticsProperty('title', title))
```
Pante, line 76:
```suggestion
```

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+      ..add(DiagnosticsProperty('initiallyExpanded', initiallyExpanded))
+      ..add(DiagnosticsProperty('child', child));
```
Pante, line 78:
```suggestion;
```

Widgets should not be added to debugFillProperties.

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+    _controller.dispose();
+    super.dispose();
+  }
+}
+
+class _Expandable extends SingleChildRenderObjectWidget {
+  final double _percentage;
+
+  const _Expandable({
+    required Widget child,
```
Pante, line 178:
```suggestion
    required super.child,
```

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+    required double percentage,
+  })  : _percentage = percentage,
+        super(child: child);
+
+  @override
+  RenderObject createRenderObject(BuildContext context) => _ExpandableBox(percentage: _percentage);
+
+  @override
+  void updateRenderObject(BuildContext context, _ExpandableBox renderObject) {
```
Pante, line 187: can use arrow syntax.

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+    renderObject.percentage = _percentage;
+  }
+}
+
+class _ExpandableBox extends RenderBox with RenderObjectWithChildMixin<RenderBox> {
+  double _percentage;
+
+  _ExpandableBox({
+    required double percentage,
+  }) : _percentage = percentage;
```
Pante, line 197: I think we can use a positional parameter here instead.

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,254 @@
...
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(DoubleProperty('percentage', percentage));
+  }
+}
+
+class _Clipper extends CustomClipper<Rect> {
+  final double percentage;
+
+  _Clipper({required this.percentage});
```
Pante, line 247: I think this can be a positional parameter too.

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,258 @@
...
+import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
+
+import 'package:forui/forui.dart';
+import 'accordion_controller_test.mocks.dart';
+
+@GenerateNiceMocks([MockSpec<AnimationController>()])
+@GenerateNiceMocks([MockSpec<Animation>()])
+void main() {
+  group('FAccordionController', () {
```
Pante, line 14: Just curious, but can the tests be simplified to use only 2 animation controllers/animations? Are there any test cases that explicitly require more than 2? Using 4 seems a little overkill.

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,258 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
+
+import 'package:forui/forui.dart';
+import 'accordion_controller_test.mocks.dart';
+
+@GenerateNiceMocks([MockSpec<AnimationController>()])
+@GenerateNiceMocks([MockSpec<Animation>()])
+void main() {
```
Pante, line 13: Most of the tests seem to be testing more than 1 non-trivial condition/behavior which is not good practice. Ideally each method under test, i.e. `toggle()` should be a group, with each condition/behavior being tested in an isolated test.

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,258 @@
...
+          when(animationControllers[i].value).thenReturn(0.0);
+          return TickerFuture.complete();
+        });
+      }
+
+      controller = FAccordionController.radio()
+        ..addListener(() {
+          count++;
+        });
+    });
```
Pante, line 171: This logic seems like it's duplicated. Please extract them out. Same with `tearDown(...)`.

forui/test/src/widgets/accordion/accordion_test.dart
```diff
@@ -0,0 +1,50 @@
+@Tags(['golden'])
```
Pante, line 1:
```suggestion
```

Don't tag non-golden tests with a golden tag.

forui/test/src/widgets/accordion/accordion_golden_test.dart
```diff
@@ -0,0 +1,78 @@
...
+library;
+
+import 'package:flutter/material.dart';
+
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:forui/forui.dart';
+import '../../test_scaffold.dart';
+
+void main() {
```
Pante, line 11: We can use `TestScaffold.app` instead, which wraps the test scaffold in a MaterialApp.

forui/test/src/widgets/accordion/accordion_test.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:forui/forui.dart';
+import '../../test_scaffold.dart';
+
+void main() {
+  group('FAccordion', () {
+    testWidgets('hit test', (tester) async {
+      var taps = 0;
+
+      await tester.pumpWidget(
+        MaterialApp(
```
Pante, line 17: Same as above.

Pante (changes requested):

forui/lib/src/widgets/accordion/accordion.dart
kawaijoe: I think it could be better to accept any widget.

forui/lib/src/widgets/accordion/accordion_controller.dart
Daviiddoo: Whats the change you wanted to apply to this?

forui/lib/src/widgets/accordion/accordion_controller.dart
Daviiddoo: Apparently this is not allowed? `invalid_internal_annotation`

Only public elements in a package's private API can be annotated as being internal.

forui/lib/src/widgets/accordion/accordion_controller.dart
Pante: I just wanted to remove the additional newline

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,363 @@
...
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
+
+import 'package:forui/forui.dart';
+import 'accordion_controller_test.mocks.dart';
+
+@GenerateNiceMocks([MockSpec<AnimationController>()])
+@GenerateNiceMocks([MockSpec<Animation>()])
```
Pante, line 12: Please add this on top of the `main` method instead.

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,363 @@
...
+void _setup(List<AnimationController> animationControllers, List<Animation<int>> animations, int length) {
+  animations.clear();
+  animationControllers.clear();
```
Pante, line 15: Why do we need to clear these? Would it be better to return a list of animation controllers and animations as a tuple instead?

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,363 @@
...
+      return TickerFuture.complete();
+    });
+  }
+}
+
+void _tearDown(List<AnimationController> animationControllers, length) {
+  for (int i = 0; i < length; i++) {
+    animationControllers[i].dispose();
+  }
+}
```
Pante, line 36: Why do we need a length for this? Can we just iterate over animation controllers?

docs/pages/docs/accordion.mdx
```diff
@@ -0,0 +1,142 @@
...
+            FAccordion(
+              controller: FAccordionController(max: 2),
+              items: [
+                FAccordionItem(
+                  title: const Text('Is it accessible?'),
+                  child: const Text('Yes. It adheres to the WAI-ARIA design pattern.'),
+                ),
+                FAccordionItem(
+                  title: const Text('Is it Styled?'),
+                  initiallyExpanded: true,
```
Pante, line 31:
```suggestion
                  initiallyExpanded: true,
                  title: const Text('Is it Styled?'),
```

docs/pages/docs/accordion.mdx
```diff
@@ -0,0 +1,142 @@
...
+            FAccordion(
+              controller: FAccordionController(),
+              items: [
+                FAccordionItem(
+                  title: const Text('Is it accessible?'),
+                  child: const Text('Yes. It adheres to the WAI-ARIA design pattern.'),
+                ),
+                FAccordionItem(
+                  title: const Text('Is it Styled?'),
+                  initiallyExpanded: true,
```
Pante, line 125:
```suggestion
                  initiallyExpanded: true,
                  title: const Text('Is it Styled?'),
```

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+import 'package:flutter_svg/svg.dart';
+import 'package:meta/meta.dart';
+
+import 'package:forui/forui.dart';
+
+/// A vertically stacked set of interactive headings that each reveal a section of content.
+///
+///
+/// See:
+/// * https://forui.dev/docs/FAccordion for working examples.
```
Pante, line 13: The link is incorrect

forui/lib/src/widgets/accordion/accordion.dart
```diff
@@ -0,0 +1,220 @@
...
+  /// See:
+  /// * [FAccordionController] for default multiple selections.
+  /// * [FAccordionController.radio] for a radio-like selection.
+  final FAccordionController? controller;
+
+  /// The items.
+  final List<FAccordionItem> children;
+
+  /// The style. Defaults to [FThemeData.accordionStyle].
+  final FAccordionStyle? style;
```
Pante, line 29:
```suggestion
  /// The style. Defaults to [FThemeData.accordionStyle].
  final FAccordionStyle? style;
  
  /// The items.
  final List<FAccordionItem> children;
```

forui/lib/src/widgets/accordion/accordion_item.dart
```diff
@@ -0,0 +1,217 @@
...
+
+  @override
+  State<FAccordionItem> createState() => _FAccordionItemState();
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties
+      ..add(DiagnosticsProperty('style', style))
+      ..add(DiagnosticsProperty('initiallyExpanded', initiallyExpanded));
```
Pante, line 45: Please use a flag property.

Pante (changes requested):

forui/test/src/widgets/accordion/accordion_controller_test.dart
```diff
@@ -0,0 +1,387 @@
...
+import 'package:flutter/widgets.dart';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
+
+import 'package:forui/forui.dart';
+import 'accordion_controller_test.mocks.dart';
+
+(List<AnimationController>, List) _setup(int length) {
```
Pante, line 11: Please type the widget.

Pante (approved):

PR #208, opened 2024-09-29
Add slider label

Pante:
**Describe the changes**
This PR builds on the previous slider PR, and introduces in-built label support.

**Checklist**


docs/pages/docs/index.mdx
```diff
@@ -35,8 +35,8 @@ To use Forui widgets in your Flutter app, import the Forui package and place the
 [`FTheme`](https://pub.dev/documentation/forui/latest/forui.theme/FTheme-class.html) widget underneath `CupertinoApp`,
 `MaterialApp`, or `WidgetsApp` at at the root of the widget tree.
 
-```dart filename="main.dart" {3,12-16}
-import 'package:flutter/material.dart';
+```dart filename="main.dart" {3,12-18}
+import 'package:flutter/widgets.dart';
```
kawaijoe, line 39: I think this example requires material

forui/example/pubspec.yaml
```diff
@@ -29,6 +29,7 @@ dependencies:
     path: ../
   intl: ^0.19.0
   sugar: ^3.1.0
+  wakelock_plus: ^1.2.8
```
kawaijoe, line 32: I don't think this should be a dependency. It should either be a `dev_dependency` or we remove it entirely.

forui/lib/src/foundation/tappable.dart
```diff
@@ -1,13 +1,36 @@
...
+    _primary = value;
+  }
+}
+
+/// The [FTappableState]'s current state.
 ///
+/// Short pressed is fired 200 ms after a tap is first detected. This is faster than long press's [kLongPressTimeout].
 @internal
-typedef FTappableState = ({bool focused, bool hovered, bool longPressed});
+typedef FTappableState = ({bool focused, bool hovered, bool shortPressed});
```
kawaijoe, line 33: Ehhh what's `shortPressed` for? Why can't we just stick with a `longPressed`?

forui/lib/src/theme/color_scheme.dart
```diff
@@ -111,6 +115,12 @@ final class FColorScheme with Diagnosticable {
     required this.border,
   });
 
+  /// Creates a disabled color from the given [color].
+  ///
+  /// See:
+  /// * [disabledColorLightness].
+  Color disable(Color color) => HSLColor.fromColor(color).withLightness(disabledColorLightness).toColor();
```
kawaijoe, line 122: Shouldn't we be changing the opacity? On a side note, if we're adding support for disabled, shouldn't this just be a color?

docs/pages/docs/slider.mdx
kawaijoe, line 1: Should we include an example for form? Similar to the other widgets.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -855,7 +855,7 @@ final class FTextField extends StatelessWidget {
       ..add(DiagnosticsProperty('undoController', undoController))
       ..add(DiagnosticsProperty('spellCheckConfiguration', spellCheckConfiguration))
       ..add(DiagnosticsProperty('suffixIcon', suffix))
-      ..add(ObjectFlagProperty.has('onSave', onSave))
+      ..add(ObjectFlagProperty.has('onSave', onSaved))
```
kawaijoe, line 858:
```suggestion
      ..add(ObjectFlagProperty.has('onSaved', onSaved))
```

kawaijoe (changes requested):

forui/example/pubspec.yaml
Pante: Nope it can't, but I think it's fine since it's the example's pubspec.

forui/lib/src/foundation/tappable.dart
Pante: I renamed it from longPress because it's technically shorter than longPress, 200ms instead of 500ms. The reason being that long press takes too long and feels unresponsive. I'm open to names though.

forui/lib/src/theme/color_scheme.dart
Pante: I think that changing the opacity isn't ideal, there was a bug with slider where changing the opacity of the active track causes the ticks/marks underneath it to be visible. 

It could be a color but that would mean we need to define all disabled colors for all styles. I'm okay with that but I feel like it should be addressed in a follow-up PR as this PR is chunky enough.

docs/pages/docs/slider.mdx
Pante: Any suggestions on how the form should look like?

forui/lib/src/theme/color_scheme.dart
Pante: Thinking about it again, I think the way forward is to declare a set of disabled colors that all widgets will use, and convert other widgets away from opacity. I got a feeling that messing with the opacity is going to cause some surprising visual oddities down the line...

PR #218, opened 2024-10-08
Replace transulcent colors with equivalent solid colors 

Pante:
**Describe the changes**
This addresses the following:
* Replace translucent colors with equivalent solid colors - using translucent colors may result in unexpected behavior when using the affected widgets against certain backgrounds.
* Introduce methods for getting equivalent solid colors of translucent colors.
* Fix inconsistent naming across widget styles.
* Refine blue screen tests.

**Checklist**


forui/lib/src/theme/color_scheme.dart
```diff
@@ -27,9 +27,6 @@ final class FColorScheme with Diagnosticable {
   /// This is typically used to determine the appearance of native UI elements such as on-screen keyboards.
   final Brightness brightness;
 
-  /// The percentage, between `0` and `1`, used to set a color's lightness to derive a disabled color.
-  final double disabledColorLightness;
```
kawaijoe, line 31: Do you think it's still nice to have this? We could have another for the hover state as well.

samples/lib/widgets/select_group.dart
```diff
@@ -73,7 +73,7 @@ class CheckboxFormState extends State<CheckboxForm> {
           mainAxisAlignment: MainAxisAlignment.center,
           crossAxisAlignment: CrossAxisAlignment.start,
           children: [
-            FSelectGroup(
+            FSelectGroup<Language>(
```
kawaijoe, line 76: Do we need to type this? Sad if we do :(

kawaijoe (commented):
Generally looks good, just a few thoughts. If we do decide to include the fields to adjust the disabled and hover opacity, I think we should include a short section under the theme page of the docs.

forui/lib/src/theme/color_scheme.dart
Pante: Ah, to make the opacity configurable?

samples/lib/widgets/select_group.dart
Pante: Yes sadly :( Dart's type inference fails here.

forui/lib/src/theme/color_scheme.dart
kawaijoe: Please add the rename to the `CHANGELOG`

samples/lib/widgets/select_group.dart
kawaijoe: :o I don't remember it throwing any compile time errors. Could you update the code sample in the docs, I don't think I have explicit generics included.

docs/pages/docs/themes.mdx
```diff
@@ -201,6 +201,14 @@ Widget build(BuildContext context) {
...
 ```
 
+#### Hovered and Disabled Colors
+
+Hovered and disabled colors are derived by adjusting the opacity. To derive these colors, use the `FColorScheme.hover`
+and `FColorScheme.disable` methods. The opacity can be adjusted with `FColorScheme.enabledHoveredOpacity` and
+`FColorScheme.disabledOpacity`.
+
+```dart filename="main.dart" {4-6}
+
```
kawaijoe, line 211: Might be a mistake here.

Daviiddoo (approved):

PR #234, opened 2024-10-20
Add responsive breakpoints

Pante:
**Describe the changes**
This PR introduces responsive breakpoints. 

~It also fixes an issue with `FTappable`'s animation playing even when it is disabled~ (The fixes have been superseded by #238).

**Checklist**


forui/lib/src/theme/breakpoints.dart
```diff
@@ -0,0 +1,146 @@
...
+  final FBreakpoint xl2;
+
+  /// Creates a [FBreakpoints].
+  const FBreakpoints({
+    this.xs = const FBreakpoint(max: 640),
+    this.sm = const FBreakpoint(min: 640, max: 768),
+    this.md = const FBreakpoint(min: 768, max: 1024),
+    this.lg = const FBreakpoint(min: 1024, max: 1280),
+    this.xl = const FBreakpoint(min: 1280, max: 1536),
+    this.xl2 = const FBreakpoint(min: 1536),
```
kawaijoe, line 73: I don't think the min and max values should be overlapped. A particular pixel size can be valid in two breakpoints which seems weird.

kawaijoe, line 73: Is it a potential issue if the user misses out on a particular breakpoint? For example, if the user needs breakpoints at xs and lg. This implementation essentially forces the user to implement, xs, sm, md, lg if they would like to scale up to lg.

The tailwind one makes sense since it only refers to the min-width and not the range of the width.

docs/pages/docs/responsive.mdx
```diff
@@ -0,0 +1,58 @@
...
+  final theme = context.theme; // FThemeData
+  final size = MediaQuery.sizeOf(context); // Size
+
+  if (theme.breakpoints.sm.contains(size)) {
+    // Do something when the sm breakpoint is active.
+    return SizedBox.square(dimension: 100);
+    
+  } else {
+    return SizedBox.square(dimension: 500);
+  }
```
kawaijoe, line 30: Might be better to convert this to a switch statement.

kawaijoe (changes requested):

kawaijoe (approved):

Daviiddoo (approved):

PR #247, opened 2024-10-27
Add `SelectTile` and `SelectTileGroup`

### LINKED ISSUE #244: SelectGroupTile

Pante:
**Describe the changes**
This PR adds `SelectTile` and `SelectTileGroup`. It also modifies `FTileGroup` to use `FLabel` internally to support descriptions and error messages.

Fixes #244.

**Checklist**


docs/pages/docs/tile/select-tile.mdx
```diff
@@ -0,0 +1,134 @@
...
+);
+```
+
+## Examples
+
+### Disabled
+
+<Tabs items={['Preview', 'Code']}>
+  <Tabs.Tab>
+    <Widget name='tile' variant='default' query={{enabled: false}}/>
```
kawaijoe, line 69:
```suggestion
    <Widget name='select-tile' variant='default' query={{enabled: false}}/>
```

forui/test/golden/select-tile-group/group/empty.png
kawaijoe, line 1: Should these be under the `zinc-dark` or `zinc-light` directory? I think this is an issue throughout the pr.

On a side note, I'm not sure if using the `zinc-dark` or `zinc-light` directory is the best way to organize since there will always be only two themes.

forui/lib/src/widgets/select_tile_group/select_tile.dart
```diff
@@ -0,0 +1,183 @@
...
+    super.key,
+  })  : _icon = prefixIcon,
+        checkedIcon = checkedIcon ?? FIcon(FAssets.icons.check),
+        // We use a transparent color instead of a sized box, to simplify calculation of the checked icon size.
+        uncheckedIcon = uncheckedIcon ?? FIcon(FAssets.icons.check, color: Colors.transparent),
+        _suffix = true;
+
+  @override
+  Widget build(BuildContext context) {
+    final FSelectTileData(:controller, :selected) = FSelectTileData.of<T>(context);
```
kawaijoe, line 108: So a `FSelectTile` must exist inside a `FSelectGroupTile`? If so, I think the docs is quite misleading as it gave the impression that a `FSelectTile` can be used stand-alone. Would it be better to just place everything under the same `FSelectGroupTile` page?

forui/lib/src/widgets/tile/tile.dart
```diff
@@ -48,12 +48,23 @@ class FTile extends StatelessWidget with FTileMixin {
   /// Provide a style to prevent inheriting from the ancestor tile group's style.
   final FTileStyle? style;
 
-  /// Whether the FTile is enabled. Defaults to true.
-  final bool enabled;
+  /// Whether the `FTile` is enabled. Defaults to true.
+  final bool? enabled;
```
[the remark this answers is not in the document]
Pante: It's because we propagate whether a tile is enabled down from the enclosing tile group. Defaulting it to `null` means "use enclosing tile group's value" while passing in a non-null value overrides the enclosing tile group's value.

forui/lib/src/widgets/select_tile_group/select_tile.dart
Pante: That's fair.

docs/pages/docs/tile/select-tile-group.mdx
```diff
@@ -0,0 +1,300 @@
...
+        key: _formKey,
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            FSelectTileGroup(
+              controller: controller,
+              label: const Text('Favorite Languages'),
+              description: const Text('Your favorite language.'),
+              validator: (values) => values?.isEmpty ?? true ? 'Please select at least one language.' : null,
```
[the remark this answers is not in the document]
Pante: Seems to be an underlying bug, select group has the same issue.

docs/pages/docs/tile/select-tile-group.mdx
```diff
@@ -0,0 +1,300 @@
...
+        key: _formKey,
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            FSelectTileGroup(
+              controller: controller,
+              label: const Text('Notifications'),
+              description: const Text('Select the notifications.'),
+              validator: (values) => values?.isEmpty ?? true ? 'Please select a value.' : null,
```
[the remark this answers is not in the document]
Pante: Seems to be an underlying bug, select group has the same issue.

forui/test/golden/select-tile-group/group/empty.png
Pante: Nope, I think there's certain golden tests validate layout which would not benefit from being tested using multiple themes. I try to avoid running such tests against multiple themes since that doesn't add much value.

I have no strong opinions but my main goal splitting them by theme was to make reviewing all images for a certain theme easier.

kawaijoe: I think that's fair, but I think in a lot of cases, these images should be tested in separate themes. For example, this image (empty.png), I think it would be nice to verify if the colors for empty work in the other theme as well. There might be a small white empty box that isn't detected because we place it against the same background.

On a side note, any thoughts on removing zinc from the theme name? Looking at it it's pretty pointless since we're only concerned about light and dark mode.

Pante: Mmm I don't mind removing zinc from theme name but it should be in another PR.

docs/pages/docs/form/checkbox.mdx
```diff
@@ -6,7 +6,7 @@ import LinkBadgeGroup from "../../../components/link-badge/link-badge-group.tsx"
 # Checkbox
 A control that allows the user to toggle between checked and not checked.
 
-For touch devices, a [switch](/docs/form/switch) is generally recommended over a checkbox.
+For touch devices, a [switch](/docs/form/switch) is generally recommended over this.
```
kawaijoe, line 9: Should this be wrapped in a callout as well?

docs/pages/docs/form/checkbox.mdx
```diff
@@ -6,12 +6,18 @@ import LinkBadgeGroup from "../../../components/link-badge/link-badge-group.tsx"
...
     <LinkBadge label="API Reference" href="https://pub.dev/documentation/forui/latest/forui.widgets.checkbox/forui.widgets.checkbox-library.html"/>
 </LinkBadgeGroup>
 
+<div className="pb-5">
+    <Callout type="info">
+        For touch devices, a [switch](/docs/form/switch) is generally recommended over this.
+    </Callout>
+</div>
+
 <div className="pb-5">
```
kawaijoe, line 21: I think you can nest both callouts together. The gap between both callouts look awfully big

PR #252, opened 2024-11-01
Add Popover Menu

Pante:
**Describe the changes**
This PR adds a popover menu. It also addresses several misc issues outlined in the CHANGELOG.md.

**Checklist**


docs/pages/docs/overlay/popover-menu.mdx
```diff
@@ -0,0 +1,133 @@
...
+                      FSelectTile(
+                        title: const Text('Grid View'),
+                        suffixIcon: FIcon(FAssets.icons.layoutGrid),
+                        value: 'Grid',
+                      ),
+                    ],
+                  ),
+                ],
+                child: FHeaderAction(
+                  icon: FIcon(FAssets.icons.ellipsis),
```
kawaijoe, line 77: Nit, but I think `ellipsis-vertical` will look better.

forui/lib/src/widgets/popover_menu.dart
```diff
@@ -0,0 +1,286 @@
...
+
+/// A [FPopoverMenuStyle]'s style.
+final class FPopoverMenuStyle extends FPopoverStyle {
+  /// The popover's default shadow in [FPopoverMenuStyle.inherit].
+  static const shadow = [
+    BoxShadow(
+      color: Color(0x0d000000),
+      offset: Offset(0, 1),
+      blurRadius: 2,
+    ),
```
kawaijoe, line 217: I noticed that we add shadows throughout all of our overlay widgets. Should this be something that we inherit?

Pante: The shadow is different from the other overlay.

kawaijoe: Should they be different?

Pante: Yup, I originally used the same shadow for popovers & popover menus. However, you mentioned that the shadows for the popover menu was too heavy which resulted in the shadows being changed.

docs/pages/docs/overlay/popover-menu.mdx
```diff
@@ -0,0 +1,122 @@
...
+                  FTileGroup(
+                    children: [
+                      FTile(
+                        prefixIcon: FIcon(FAssets.icons.list),
+                        title: const Text('List View'),
+                      ),
+                      FTile(
+                        prefixIcon: FIcon(FAssets.icons.layoutGrid),
+                        title: const Text('Grid View'),
+                      ),
```
kawaijoe, line 67: I think we should `onPress` on these too.

forui/lib/src/widgets/popover_menu.dart
kawaijoe: Ah I see. If that's the case, would it make sense for both of them to adopt the lighter shadow? Imo, less is better than having too much. In addition, there's no gurantee that a popover is larger than a popover menu.

The popover menu is actually a good example that a popover can be small.

Pante: Ehh sure, no strong opinion on it.

kawaijoe (approved):

PR #338, opened 2025-01-04
Add properties

Pante:
**Describe the changes**

**Checklist**
- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md).


forui/lib/src/foundation/form_field_properties.dart
```diff
@@ -46,53 +45,36 @@ abstract class FFormField<T> extends StatelessWidget {
...
+  final Widget Function(BuildContext, String) errorBuilder;
+
+  /// Creates a [FFormFieldProperties].
+  const FFormFieldProperties({
+    this.onSaved,
     this.forceErrorText,
     this.validator,
     this.enabled = true,
     this.autovalidateMode = AutovalidateMode.disabled,
-    this.restorationId,
```
kawaijoe, line 68: Just curious why is this removed?

forui/lib/src/foundation/focusable_properties.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+
+/// Properties for focusable widgets.
+class FFocusableProperties with Diagnosticable {
```
kawaijoe, line 5: Should properties be mixins? Feels like it can be useful to add helper methods which widgets can use.

kawaijoe (commented):
Generally lgtm, but i think finding the right level of abstraction would be the interesting portion.

PR #450, opened 2025-03-20
Create `FSelect`

### LINKED ISSUE #162: Select Component
Implement a selection/dropdown/combo box style component. This is a key component in forms/settings pages.

**Additional context**
Shadcn implementation: https://ui.shadcn.com/docs/components/select

Pante:
**Describe the changes**
Resolves #162.

**Checklist**


forui/lib/src/widgets/select/select_item.dart
```diff
@@ -0,0 +1,291 @@
...
+  /// The value.
+  final T value;
+
+  /// True if the item is enabled. Disabled items cannot be selected, and is skipped during traversal.
+  ///
+  /// Defaults to the value inherited from the parent [FSelectSection] or [FSelect].
+  final bool? enabled;
+
+  /// The title.
+  final Widget title;
```
kawaijoe, line 134: I think title is a bit inconsistent here, it should either be named `label` or `child`. I think if we're trying to move towards using `child`, then that should be preferred.

forui/lib/src/widgets/select/select_item.dart
```diff
@@ -0,0 +1,291 @@
...
+  /// The style. Defaults to the [FSelectSectionStyle] inherited from the parent [FSelect].
+  final FSelectSectionStyle? style;
+
+  /// True if the section is enabled. Disabled sections cannot be selected, and is skipped during traversal.
+  ///
+  /// Defaults to inheriting from the [FSelect].
+  final bool? enabled;
+
+  /// The title.
+  final Widget title;
```
kawaijoe, line 27: Similar issue here.

forui/lib/src/widgets/select/select.dart
```diff
@@ -0,0 +1,566 @@
...
+  final Widget? description;
+
+  /// {@macro forui.foundation.form_field_properties.errorBuilder}
+  final Widget Function(BuildContext, String) errorBuilder;
+
+  /// {@macro forui.foundation.form_field_properties.enabled}
+  final bool enabled;
+
+  /// {@macro forui.foundation.form_field_properties.onSaved}
+  final FormFieldSetter<T>? onSaved;
```
kawaijoe, line 86: I just noticed this. Should this be in the past tense?

Pante: We name it `onSaved` because that's what Flutter names theirs. We could name it `onSave` but this would make it inconsistent with other Forui widgets that extend Flutter's `FormField` where it is named `onSaved`.

I think this ties in with me wanting to rethink how we name our callbacks. 

Flutter is inconsistent in naming fields. It's a mistake that they've acknowledged/regret but can't fix due to breaking changes, https://github.com/flutter/flutter/issues/24722#issuecomment-2738138585.

When we first created Forui, we wanted to respect Flutter's style guide, and use present tense for everything. This is also acknowledge in our own contributing guide. Although sound in theory, cracks began to start forming when we slowly incorporated Flutter's widgets/classes, i.e. FormField. This led to a weird mix where most of our fields are named using present tense except when interfacing with Flutter's core widgets. 

I guess the question is whether being inconsistent with Flutter's inconsistencies making things even more inconsistent :p

forui/lib/src/widgets/select/select_item.dart
```diff
@@ -0,0 +1,291 @@
...
+  @override
+  final TextStyle enabledTitleTextStyle;
+
+  /// The disabled title's text style.
+  @override
+  final TextStyle disabledTitleTextStyle;
+
+  /// The padding around the title. Defaults to `EdgeInsetsDirectional.only(start: 15, top: 7.5, bottom: 7.5, end: 10)`.
+  @override
+  final EdgeInsetsGeometry titlePadding;
```
kawaijoe, line 91: Naming here aren't updated.

forui/lib/src/widgets/select/select.dart
kawaijoe: Hmmm why not just rename the field `onSave`?

I feel like following flutter's naming convention is just a potential rabbit hole. For example, it is possible that certain names will be created on Forui before a comparable implementation on Flutter. Does that mean that we should break Forui just so that it matches with Flutter's convention?

My take is that we scrub any inconsistent naming from the Flutter library and manage the naming on our end.

kawaijoe (commented):
One additional thing I noticed is that the scroll bar is really hard to see in the dark mode. Is that a theme setting that can be adjusted?

forui/lib/src/widgets/select/select_item.dart
```diff
@@ -0,0 +1,291 @@
...
+  final Widget child;
+
+  /// The icon displayed when the item is selected. Defaults to a check icon.
+  final Widget selectedIcon;
+
+  /// Creates a [FSelectItem].
+  FSelectItem({required this.value, required this.child, this.style, this.enabled, Widget? icon, super.key})
+    : selectedIcon = icon ?? FIcon(FAssets.icons.check);
+
+  /// Creates a [FSelectItem] that displays the [value] as its title.
```
kawaijoe, line 143: Outdated comment.

kawaijoe (commented):
Might want to do a sweep of the select files. Still finding outdated references.

PR #472, opened 2025-04-03
Add lerping & renaming of parameters

### LINKED ISSUE #468: More concise naming for `FThemeData` fields
Sorry if this sounds nitpicking, but I often try to use values from `context.theme` for styling my other widgets, and occasionally got a bit annoyed by how verbose the field names are.

For example, I feel like naming things like `context.theme.text.sm` and `context.theme.color.muted` would be more intuitive and nicer to see than `context.theme.typography.sm` and `context.theme.colorScheme.muted`. They would also be more similar to the class names in shadcn like `text-sm` and `bg-muted`.

Curious whether there are certain considerations to use their verbose name based on the class name instead of the simplified name? (perhaps do you plan to introduce sibling fields with different suffix?)

Pante:
**Describe the changes**
Fixes #468.

**Checklist**


kawaijoe (commented):
Generally looks good but I think we should include some documentation on the website for this. Potentially break up the theme section of the docs or just add a section.

PR #497, opened 2025-04-25
Update FHeader.nested ( Fix #496 )

### LINKED ISSUE #496: Add alignment to Header's title.
**Is your feature request related to a problem? Please describe.**
Currently, the header title, which is the primary text or widget displayed in the header bar, is always positioned in the center by default and there is no direct parameter to control its horizontal alignment. 

This limitation makes it difficult to implement design patterns where the title needs to be aligned to the start or end of the header,.

**Describe the solution you'd like**
This property should accept an alignment value, ideally using a standard Flutter type like AlignmentGeometry. 

This would allow developers to easily specify the horizontal alignment of the title, choosing from options such as Alignment.centerStart, Alignment.center, or Alignment.centerEnd. 

Implementing this parameter would enable direct control over the title's placement within the header layout.

I will make a PR if requested.

a-man-called-q:
**Describe the changes**
Add `TextAlign titleAlignment,` for `FHeader.nested`.

In the process, I have to change how `FHeader.nested` structure the widget.
 
This supposed to fix #496 
  
**Checklist**


Pante:
@kawaijoe Is there a reason why we used a stack in the original implementation? I remember there was a reason but I can't recall it off the top of my head.

forui/CHANGELOG.md
```diff
@@ -84,6 +84,8 @@ Several minor tweaks have been made to `FHeader` to improve usability and consis
 
 * Change `FHeader(title: ...)` to be optional.
 * Change `FHeader.nested(title: ...)` to be optional.
+* Add optional `titleAlignment` to `FHeader.nested`.
+* Refactor `FHeader.nested` so it would support `titleAlignment`
```
Pante, line 88:
```suggestion
* Add `titleAlignment` to `FHeader.nested(...)`.
```

samples/lib/widgets/header.dart
```diff
@@ -26,7 +26,8 @@ class NestedHeaderPage extends Sample {
   @override
   Widget sample(BuildContext context) => FHeader.nested(
     title: const Text('Appointment'),
-    prefixes: [FHeaderAction.back(onPress: () {})],
+    titleAlignment: TextAlign.start,
+    prefixes: [FHeaderAction.back(onPress: () {}), FHeaderAction(icon: const Icon(FIcons.plus), onPress: () {})],
```
Pante, line 30: I don't think we should the sample in this case.

forui/lib/src/widgets/header/nested_header.dart
```diff
@@ -83,6 +88,8 @@ final class _FNestedHeader extends FHeader {
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
-    properties.add(DiagnosticsProperty('style', style));
+    properties
+      ..add(DiagnosticsProperty('style', style))
+      ..add(DiagnosticsProperty<TextAlign>('titleAlignment', titleAlignment));
```
Pante, line 93:
```suggestion
      ..add(EnumProperty('titleAlignment', titleAlignment));
```

forui/lib/src/widgets/header/nested_header.dart
```diff
@@ -22,27 +22,29 @@ final class _FNestedHeader extends FHeader {
...
 
+  final TextAlign titleAlignment;
+
   /// Creates a [_FNestedHeader].
   const _FNestedHeader({
     this.style,
     this.prefixes = const [],
     this.suffixes = const [],
     super.title = const SizedBox(),
+    this.titleAlignment = TextAlign.center,
```
Pante, line 33:
```suggestion
    this.titleAlignment = TextAlign.center,
    super.title = const SizedBox(),
```

Pante (changes requested):

kawaijoe:
> @kawaijoe Is there a reason why we used a stack in the original implementation? I remember there was a reason but I can't recall it off the top of my head.

It's probably something to do with centering. Using a row could result in the title being off centered if the icon on the left or right is missing/different sizes.

a-man-called-q:
@kawaijoe agreed.. now that you said it, it does look annoying.

How about if the title is centered, use align and if not... use expanded.

Pante:
I think that could potentially work, but we'll want to introduce golden tests to make sure we aren't missing anything.

Specifically we will probably want to test:
* Prefix + title align to start + no suffix
* Prefix + title align to end + no suffix

forui/lib/src/widgets/header/nested_header.dart
a-man-called-q: image

You want me to cast it to Enum? instead ?

Pante: Whoops, my bad. I keep forgetting that it's not an enum.

a-man-called-q: Ah lol ok. I'll commit it soon.

forui/lib/src/widgets/header/nested_header.dart
```diff
@@ -53,36 +58,64 @@ final class _FNestedHeader extends FHeader {
...
+    );
+  }
+
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
-    properties.add(DiagnosticsProperty('style', style));
+    properties
+      ..add(DiagnosticsProperty('style', style))
+      ..add(DiagnosticsProperty<AlignmentGeometry>('titleAlignment', titleAlignment));
```
Pante, line 119:
```suggestion
      ..add(DiagnosticsProperty('titleAlignment', titleAlignment));
```

Pante (changes requested):
Generally looks good to me! I think the only minor nit is I think we should apply actionSpacing between the title and prefix/suffix when it is aligned to the start/end.

Pante (approved):
LGTM! Thanks for the PR.

PR #499, opened 2025-04-26
Fix dialog not handling infinitely sized body correctly

### LINKED ISSUE #498: `FDialog` fails to handle size constraint and crashes
To reproduce, try running this sample code:

```dart
showDialog(
  context: context,
  builder:
      (context) => FDialog(
        direction: Axis.horizontal,
        body: Container(
          constraints: const BoxConstraints(
            maxHeight: 300,
            maxWidth: 300,
          ),
          child: SingleChildScrollView(
            child: Text.rich(
              WidgetSpan(
                child: Stack(
                  children: [
                    Container(
                      height: 200,
                      width: double.infinity,
                      color: Colors.red,
                    ),
                  ],
                ),
              ),
            ),
          ),
        ),
        actions: const [],
      ),
);
```

I ended up getting the following error:

```
════════ Exception caught by rendering library ═════════════════════════════════
The following assertion was thrown during performLayout():
Assertion failed: file:///[redacted]/flutter/packages/flutter/lib/src/rendering/stack.dart:664:12
size.isFinite
is not true

The relevant error-causing widget was:
    FDialog FDialog:file:///[redacted]

When the exception was thrown, this was the stack:
../dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 266:3     throw_
errors.dart:266
../dart-sdk/lib/_internal/js_dev_runtime/private/profile.dart 117:39               assertFailed
profile.dart:117
package:flutter/src/rendering/stack.dart 664:16                                    [_computeSize]
stack.dart:664
package:flutter/src/rendering/stack.dart 621:12                                    computeDryLayout
stack.dart:621
package:flutter/src/rendering/box.dart 2030:25                                     [_computeDryLayout]
box.dart:2030
package:flutter/src/rendering/box.dart 1054:21                                     <fn>
```

Basically, the `RenderStack` object fails to detect the width constraint up in the tree and crashes due to infinite width.

For comparison, try using the official `Dialog`:

```dart
showDialog(
  context: context,
  builder:
      (context) => Dialog(
        child: Container(
          constraints: const BoxConstraints(
            maxHeight: 300,
            maxWidth: 300,
          ),
          child: SingleChildScrollView(
            child: Text.rich(
              WidgetSpan(
                child: Stack(
                  children: [
                    Container(
                      height: 200,
                      width: double.infinity,
                      color: Colors.red,
                    ),
                  ],
                ),
              ),
            ),
          ),
        ),
      ),
);
```

This works fine as expected.

Pante:
**Describe the changes**
Fixes #498 

**Checklist**

### DIFF
--- a/forui/lib/src/widgets/dialog/dialog_content.dart
+++ b/forui/lib/src/widgets/dialog/dialog_content.dart
@@ -29,33 +29,31 @@ sealed class Content extends StatelessWidget {
   });
 
   @override
-  Widget build(BuildContext context) => IntrinsicWidth(
-    child: Padding(
-      padding: style.padding,
-      child: Column(
-        mainAxisSize: MainAxisSize.min,
-        crossAxisAlignment: alignment,
-        children: [
-          if (title case final title?)
-            Padding(
-              padding: const EdgeInsets.only(bottom: 4),
-              child: Semantics(
-                container: true,
-                child: DefaultTextStyle.merge(textAlign: titleTextAlign, style: style.titleTextStyle, child: title),
-              ),
+  Widget build(BuildContext context) => Padding(
+    padding: style.padding,
+    child: Column(
+      mainAxisSize: MainAxisSize.min,
+      crossAxisAlignment: alignment,
+      children: [
+        if (title case final title?)
+          Padding(
+            padding: const EdgeInsets.only(bottom: 4),
+            child: Semantics(
+              container: true,
+              child: DefaultTextStyle.merge(textAlign: titleTextAlign, style: style.titleTextStyle, child: title),
             ),
-          if (body case final body?)
-            Padding(
-              padding: const EdgeInsets.only(bottom: 4),
-              child: Semantics(
-                container: true,
-                child: DefaultTextStyle.merge(textAlign: bodyTextAlign, style: style.bodyTextStyle, child: body),
-              ),
+          ),
+        if (body case final body?)
+          Padding(
+            padding: const EdgeInsets.only(bottom: 4),
+            child: Semantics(
+              container: true,
+              child: DefaultTextStyle.merge(textAlign: bodyTextAlign, style: style.bodyTextStyle, child: body),
             ),
-          if (title != null && body != null) const SizedBox(height: 8),
-          _actions(context),
-        ],
-      ),
+          ),
+        if (title != null && body != null) const SizedBox(height: 8),
+        _actions(context),
+      ],
     ),
   );
 
--- a/forui/CHANGELOG.md
+++ b/forui/CHANGELOG.md
@@ -193,7 +193,8 @@ have been renamed to `semanticsLabel` for consistency.
 
 * Replace `FSelectGroupItem` with `FCheckbox.grouped(...)` and `FRadio.grouped(...)`.
 
-* Fix an issue where header spacing appeared in incorrect order.
+* Fix `FDialog` not handling infinitely sized body correctly.
+* Fix `FHeader` spacing appearing in incorrect order.
 * Fix `FResizable` not guarding against precision errors in assertions.
 * Fix `FSelectGroup` not setting its `FormField`'s initial value.
 * Fix `FSelectTileGroup` not setting its `FormField`'s initial value.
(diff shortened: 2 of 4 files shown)


forui/lib/src/widgets/dialog/dialog_content.dart
```diff
@@ -29,33 +29,31 @@ sealed class Content extends StatelessWidget {
   });
 
   @override
-  Widget build(BuildContext context) => IntrinsicWidth(
```
santiagos01, line 32: Thank you @Pante for helping to fix this!

If I understand correctly, `IntrinsicWidth` seems to be the culprit here. I wonder whether it makes sense to remove the expectation of using `IntrinsicWidth` / `IntrinsicHeight` across some other parts? For instance, `FButton` now needs to always be wrapped using `IntrinsicWidth` if one doesn't want it to expand to the maximum available width (which I think may be a more common expectation).

side note: as [mentioned](https://api.flutter.dev/flutter/widgets/IntrinsicWidth-class.html) by the flutter team, intrinsic widgets are also more expensive (O(N^2)) because of their multi-pass layout.

Pante: I think we’ll need to reevaluate in a case by case basis. For FButton, we recently introduced a instrinicWidth param that sets the main axis size of an internal row.

santiagos01: > For `FButton`, we recently introduced a `instrinicWidth` argument that sets the main axis size of an internal row. It's turned off by default since initial testing across our internal widgets seem to indicate that there are more cases where we want it to expand than not.

I see, thanks for the clarification. And just noticed the `intrinsicWidth` field, sounds good will just use it then (guess I can create my own wrapper if I want to change the default and reduce code duplication later).

> Would really appreciate more data on this outside of Forui.

Hmm, it probably depends on the target platform and subjective design choices. For mobile, yes I guess it's pretty common to have the button expanded given the small screen/viewport width (well, except those foldable phones). But for tablet, desktop, and web, given the larger width typically available, I think expanding buttons is less common...

> Do you have other widgets in mind?

Hmm, I think my primary concern is only around `FButton`, which I uses the most at the moment.

But some quick code search gives me this:
- [Badge](https://github.com/forus-labs/forui/blob/e246f909baf2b4f31eb815268f436cd3e7ee8d1d/forui/lib/src/widgets/badge/badge.dart#L43): maybe add an option to disable intrinsic as well? since I think it wouldn't be useful unless one displays a list of badges? 
- To improve performance, maybe can consider to use the `intrinsicWidth` field when [building dialog action buttons](https://github.com/forus-labs/forui/blob/e246f909baf2b4f31eb815268f436cd3e7ee8d1d/forui/lib/src/widgets/dialog/dialog_content.dart#L88)?

Pante: Hmmm, I guess we can default `intrinsinicWidth` to false for now. 

> edit: Just checked the impl for intrinsicWidth, looks like the name may be a bit misleading since it doesn't actually use any intrinsic widget (which is good!). Do you think it would be better to rename it into something like expand/expanded?

We opted not to name it `expands`/`expanded` as `TextField` has a `expands` field which affects it height.

> [Badge](https://github.com/forus-labs/forui/blob/e246f909baf2b4f31eb815268f436cd3e7ee8d1d/forui/lib/src/widgets/badge/badge.dart#L43): maybe add an option to disable intrinsic as well? since I think it wouldn't be useful unless one displays a list of badges?

We'll probably look into this in the next update as we got other things planned for badge.

> To improve performance, maybe can consider to use the intrinsicWidth field when [building dialog action buttons](https://github.com/forus-labs/forui/blob/e246f909baf2b4f31eb815268f436cd3e7ee8d1d/forui/lib/src/widgets/dialog/dialog_content.dart#L88)?

That's a good catch!

santiagos01: > Hmmm, I guess we can default intrinsinicWidth to false for now.

Isn't the default already `false` now? or you mean `true` here?

> We opted not to name it expands/expanded as TextField has a expands field which affects it height.

I see. Well, maybe can consider something like `expandWidth` / etc., but no strong opinion here.

> We'll probably look into this in the next update as we got other things planned for badge.

Thanks!

Pante: > Isn't the default already false now? or you mean true here?

Ah, I meant keeping it as false for now.

PR #529, opened 2025-05-11
Sidebar for `FScaffold`

kawaijoe:
**Describe the changes**

**Checklist**


forui/lib/src/widgets/scaffold.dart
```diff
@@ -27,6 +27,9 @@ class FScaffold extends StatelessWidget {
   /// The optional header displayed at the top of the scaffold.
   final Widget? header;
 
+  /// The optional sidebar displayed at the left of the scaffold.
```
Pante, line 30: Curious, but should this always be on the LHS? I'm wondering if the navbar should move to the RHS in RTL locales.

Pante: I think we should also document how the footer, header & sidebar interact.

Pante (changes requested):

forui/lib/src/widgets/scaffold.dart
kawaijoe: Yep, it does get flipped. I think the existing code handles that automatically through `Row`, and I've updated the docs.

For reference:
https://m3.material.io/foundations/layout/understanding-layout/bidirectionality-rtl

Pante (approved):
LGTM

PR #600, opened 2025-06-30
Add item

Pante:
**Describe the changes**

**Checklist**
- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md).
- [ ] I have included the relevant unit/golden tests.
- [ ] I have included the relevant samples.
- [ ] I have updated the documentation accordingly.
- [ ] I have added the required flutters_hook for widget controllers.
- [ ] I have updated the [CHANGELOG.md](../forui/CHANGELOG.md) if necessary.

PR #604, opened 2025-07-03
Add item group

Pante:
**Describe the changes**
Follow-up to #600 that simplifies the implementation & adds support for FItemGroup.

**Checklist**


docs/app/docs/data/item-group/page.mdx
```diff
@@ -0,0 +1,291 @@
+import {Callout} from "nextra/components";
```
kawaijoe, line 1: I think it generally looks good, but I would argue that we probably don't want users using `Item` and `ItemGroup` by itself. It could probably be better categorized as a foundation widget.

What are your thoughts? @Pante

Issue #601, opened 2025-06-30
No SelectContentData found in context. Try explicitly specifying FSelect's type parameter, i.e. FSelect<String>.


helloalbin:
**Describe the bug**
I am getting the following error, if the FSelect type is any object other than String

```
No SelectContentData found in context. Try explicitly specifying FSelect's type parameter, i.e. FSelect<String>.
'package:forui/src/widgets/select/content/content.dart':
Failed assertion: line 30 pos 7: 'result != null'
```

**To Reproduce**

Following code throws this error

```
FSelect<MenuModel>(
      controller: _controller,
   //   style: FSelectStyle(...),
      label: const Text('Country'),
      description: const Text('Select your country of residence'),
      hint: 'Choose a country',
      format: (value) => value.title!.toUpperCase(),
      onChange: (value) => print('Selected country: $value'),
      onSaved: (value) => print('Saved country: $value'),
   //   style: FSelectStyle.inherit(...),
      autovalidateMode: AutovalidateMode.onUserInteraction,
      builder: (context, styles, child) => child!,
      prefixBuilder: (context, styles, _) => Icon(FIcons.globe),
      suffixBuilder: (context, styles, _) => Icon(FIcons.arrowDown),
      popoverConstraints: const FAutoWidthPortalConstraints(maxHeight: 400),
      clearable: true,
      contentScrollHandles: true,
      children: [
        FSelectSection(
          label: const Text('North American Countries'),
          children: [
            FSelectItem.from(
              child: const Text('United States'),
              value: 'us',
            ),
            FSelectItem(
              'Canada',
              'ca',
            ),
          ],
        ),
        FSelectItem(
          'Japan',
           'jp',
        ),
      ],
    );
```
**Expected behavior**
The Select menu should be displayed.

**Device (please complete the following information):**
 - Flutter version:  3.32.4
 - Forui version: 0.12.0
 - Device: iPhone emulator

**Additional context**
Works fine if the type is String

Pante:
You need to change `FSelectItem.value` to the correct type (`MenuModel`), it's currently a `string`.

helloalbin:
Thank you. It's working after adding a type to FSelectItem. I'll recommend adding a note to our documentation to clarify this behavior for future users.

Pante:
I'm gonna reopen this issue so that I remember to add a note in the docs & update the error message.

Issue #469, opened 2025-04-02
`FPopoverMenu` default style


santiagos01:
Not sure whether people normally would notice this, but I feel that the current default style with "partial divider" between items in the same group makes it look a bit off on hover (as if the part without divider around it has 1px less height top & bottom).

![Image](https://github.com/user-attachments/assets/208931e4-6e24-4864-be75-27b135e9475d)

Would it be better to style the current popover menu closer to the one by shadcn? (which is also similar to how almost all other apps do it?)

![Image](https://github.com/user-attachments/assets/8b143b8a-72d0-414d-9738-f37d52363a62)

Pante:
> Is there any particular reason that the team prefers to have partial divider between items within the same group? (perhaps there's any particular design convention that you try to follow?)

Our main inspiration was iOS/Cuptertino for this. We chose partial dividers as we wanted a distinct way to separate two different tiles such that it is easier for users to tap on the correct tile on touch devices. That said, dividers can easily be changed via [https://pub.dev/documentation/forui/latest/forui.widgets.tile/FTileDivider.html].

> Hmm, is there a way to use one same class for both desktop and mobile? Or I guess technically one can create a wrapper class to do the type switch as well...

This I'm not too sure, we'll need investigate further before anything concrete can be said.

PR #605, opened 2025-07-05
Misc improvements

### LINKED ISSUE #603: Allow custom FTappable keyboard shortcut for onPress
**Is your feature request related to a problem? Please describe.**
I have a modal that shows a list of items using `FTileGroup` and an `FItem` for each entry. Tappable has a built in shortcut so that if I specify an onPress handler for an item, and press enter when that item has focus, the onPress is triggered. However, I prefer to use the enter key to submit the form (same as pressing the `OK` button) and use the space bar for triggering onPress via keyboard.

**Describe the solution you'd like**
One solution would be to allow customising which key triggers `onPress` as an alternative to `enter`

**Describe alternatives you've considered**
I was able to work around the problem by creating my own tile and `FTileMixin` however this seems like a heavy-handed approach.

**Additional context**
In this screenshot, the goal is to allow toggling items with space, while the enter key closes the modal.

Image

Pante:
**Describe the changes**
Prepares for 0.13.0 & fixes #603, #601, #469.

It introduces a tile & item variant for `FPopoverMenu` and `FBreadcrumbItem.collapsed`.

**Checklist**


kawaijoe (commented):
Generally lgtm! The only thing would be that the padding looks slightly strange on popover.

Image

On a side note, could you add the callout for item as well? I saw that it is added for item group.

Issue #615, opened 2025-07-09
Nested Header title vertically not centered


luishugoooo:
**Describe the bug**
See title

**To Reproduce**
```
FHeader.nested(
                titleAlignment: Alignment.center,
                title: Center(child: Text("Text")),
                prefixes: [
                  FHeaderAction(
                    icon: Icon(FIcons.menu),
                    onPress: () {
                      showFSheet(
                        context: context,
                        builder: (context) => OpenShelfSidebar(
                          children: children,
                          onChange: onChange,
                          currentIndex: index,
                        ),
                        side: FLayout.ltr,
                      );
                    },
                  ),
                ],
              )
```

**Expected behavior**
Title is centered

**Screenshots**

<img width="447" height="35" alt="Image" src="https://github.com/user-attachments/assets/f307cf2d-7be5-4b56-90f3-fbc79145d24a" />

**Device (please complete the following information):**
 - Flutter version: latest stable
 - Forui version: 0.13.1
 - Device: MacOS, iPhone, Android

**Additional context**
-

PR #661, opened 2025-08-18
🐞 fix: Nested Header title vertically not centered #615

lm83680:
**Describe the changes**
- 🐞 fix: Nested Header title vertically not centered [Issue #615](https://github.com/forus-labs/forui/issues/615)
### Before 
Before 
### After
After 

**Checklist**


forui/lib/src/widgets/header/nested_header.dart
```diff
@@ -65,23 +65,33 @@ class _FNestedHeader extends FHeader {
...
-                prefixes,
-                Expanded(child: title),
-                suffixes,
-              ],
-            );
+      // Layout title with prefixes/suffixes based on alignment.
+      // - Horizontal center (x == 0) uses Stack for perfect centering.
+      // - Other horizontal positions use Row with vertical alignment determined by y.
+      if (alignment.x == 0 || ![-1, 0, 1].contains(alignment.y)) {
+        title = IntrinsicHeight(
```
Pante, line 72: Can you explain why the stack needs to be wrapped in a `IntrinsicHeight`?

lm83680: .Thanks for the review! Let me clarify.

When either prefixes or suffixes has a greater height than the title, the title cannot perceive the true height of the Stack. In special alignments (e.g. Center, Alignment(0.5, 0.5)), this causes the title to be positioned incorrectly.

I actually tried to avoid using IntrinsicHeight. but i can't

lm83680: @Pante 
The purpose of IntrinsicHeight is:
When the height of child widgets is uncertain, it forces the parent to size itself based on the intrinsic height of its children.

By default, a Stack does not expand to the maximum height of its children; it only lays out according to the positioning rules.

When either prefixes or suffixes is taller than the title, the Stack's height is underestimated, causing the title to be positioned incorrectly when using center alignment or Alignment(0.5, 0.5).

Wrapping with IntrinsicHeight allows the parent to measure the intrinsic height of its children, so the Stack becomes aware of the true height range, and the title can be accurately centered.

Completely avoiding IntrinsicHeight is quite difficult, because the core issue is:

Stack by default does not expand to the maximum height of its children; it only lays out according to the positioning of its widgets.

Your title needs to be centered or aligned based on the actual height of prefixes/suffixes, but the Stack itself does not “perceive” the height of these child widgets.

So the purpose of IntrinsicHeight here is to allow the Stack to perceive the true height of prefixes/suffixes and prevent the title from being misaligned due to underestimated height.

Pante: I think you really raised a good point about the prefixes & suffixes being larger than the title which is something we didn't consider. 

However, maybe I'm understanding something, does the PR fix the prefixes & suffixes not being vertically aligned when they are smaller than the title? 

Asking because we had an internal discuss regarding the issue awhile back, and we felt like that most fool-proof way around it is to create a `RenderBox` with a custom layout strategy.

lm83680: Thank you for the reminder. Through some testing, I found that the current PR basically meets common layout needs. image
However, during testing I found one scenario: when the user cannot determine the height of the prefixes, suffixes, or title, it is impossible to satisfy both of the following at the same time:

1. The title is not vertically centered.
2. The prefixes or suffixes remain always centered.

This is my test case.

```dart

class Sandbox extends StatefulWidget {
  const Sandbox({super.key});

  @override
  State<Sandbox> createState() => _SandboxState();
}

class _SandboxState extends State<Sandbox> with SingleTickerProviderStateMixin {
  late final FContinuousSliderController c;
  double value = 0;
  @override
  void initState() {
    super.initState();
    c = FContinuousSliderController(selection: FSliderSelection(max: 0.1));
  }

  @override
  void dispose() {
    c.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    return Column(
      spacing: 5,
      children: [
        FSlider(
          controller: c,
          onChange: (v) => setState(() {
            value = (v.offset.max * 100).clamp(20, 100);
          }),
        ),
        FHeader.nested(
          title: Text('$value', style: TextStyle(fontSize: 40)),
          prefixes: [FHeaderAction(icon: Icon(FIcons.heading, size: value), onPress: null)],
        ),
        FLabel(
          axis: Axis.vertical,
          label: Text('Title not vertical center'),
          child: FHeader.nested(
            titleAlignment: AlignmentGeometry.bottomLeft,
            title: Text('$value', style: TextStyle(fontSize: 40)),
            prefixes: [FHeaderAction(icon: Icon(FIcons.heading, size: value), onPress: null)],
          ),
        ),
        FLabel(
          axis: Axis.vertical,
          label: Text('Unless he can confirm the height of the title'),
          child: FHeader.nested(
            titleAlignment: AlignmentGeometry.bottomLeft,
            title: Text('$value', style: TextStyle(fontSize: 40)),
            prefixes: [
              ConstrainedBox(
                constraints: BoxConstraints(minHeight: 40),
                child: Center(
                  child: FHeaderAction(icon: Icon(FIcons.heading, size: value), onPress: null),
                ),
              ),
            ],
          ),
        ),
      ],
    );
  }
}
```
Unless he can confirm the title because he is in the 'Row' branch and not 'Stack'.
<video src="https://github.com/user-attachments/assets/8e8d145a-e702-4f53-af70-b9e90bf03957" controls width="600"></video>

Pante: Hmm interesting, what do you think @kawaijoe ?

Pante:
@lm83680 Sorry for the long reply, we've discussed this internally and I think we'll be going forward with a different implementation that uses a custom RenderBox as there will be more robust and handle most edge cases better.

Issue #668, opened 2025-08-23
Add reveal/hide toggle for FTextField.password


cacing69:
Currently, the FTextField.password widget only accepts input in an obscured state.

```dart
FTextField.password(
  label: Text('Password'),
  hint: 'john@doe.com',
),
```

Expected behavior:
The password field should have a toggle icon/button (e.g., an eye icon) to reveal or hide the password text by switching obscureText from true to false.

Why this is needed:
This is a common UX pattern for password fields, helping users verify their input without compromising security.

Pante:
I personally think this one's best left to the user. I'm not too sure how this will work with a confirm password field where clicking on the icon of one field should obscure/un-obscure both fields.

iamutaki:
I personally think it be powerful to have the reveal option included in the password field. The overall experience and control can really depend on the developer’s needs.

I respect whatever decision you make, just sharing what I think could work best.

Pante:
I'm not opposed to it, it's just that I can't imagine how the API will look like. Hmmm....

Pante:
I'm going to re-open this issue since thinking about it again, perhaps we can do something about this.

iamutaki:
Great to see this open again! I still think adding a toggle would be really valuable, and I’m excited to see how it progresses.

MoathCodes:
I wanted to play around and see what i can do, hopefully this is a good starting point, I've added this to the API:
```dart
/// Builds a custom obscure-text toggle control.
typedef FTextFieldObscureToggleBuilder = Widget Function(
  BuildContext context,
  FTextFieldStyle style,
  Set<WidgetState> states,
  bool obscureText,
  VoidCallback toggle,
);

class FTextField extends StatefulWidget {
  /// If true, shows a built-in visibility toggle when [obscureText] is true
  /// and the field is enabled. Defaults to false.
  final bool showObscureToggle;

  /// Custom builder for the visibility toggle. If null, a default ghost
  /// icon button is used.
  final FTextFieldObscureToggleBuilder? obscureToggleBuilder;

  // Constructors:
  const FTextField({ this.showObscureToggle = false, this.obscureToggleBuilder, ... });
  const FTextField.email({ this.showObscureToggle = false, this.obscureToggleBuilder, ... });
  const FTextField.password({
    this.showObscureToggle = true, // enabled by default here
    this.obscureToggleBuilder,
    ...
  });
}
```
[...]

PR #679, opened 2025-08-30
Adding a visibility toggle for FTextField.password (#668)

MoathCodes:
**Describe the changes**
This PR is from #668, these changes add new visibility toggle for the `FTextField.password`, I still need to change the `FTextFormField.password` as well, **however while implementing this I realized that because I've made the TextField into a static method the parameters no longer have documentations**.
I think I've missed some better architectural decisions, please tell me what do you think.
###  Usage Examples

```dart
// ✅ Basic usage - automatic toggle and controller
FTextField.password()

// ✅ Custom toggle - requires explicit controller
final controller = ValueNotifier<bool>(false);
FTextField.password(
  visibilityController: controller,
  toggleBuilder: (context, style, states) => IconButton(
    onPressed: () => controller.value = !controller.value,
    icon: Icon(controller.value ? Icons.visibility_off : Icons.visibility),
  ),
)

// ✅ No toggle - password always obscured  
FTextField.password(toggleBuilder: null)
```
**Checklist**


forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -37,6 +37,17 @@ typedef FFieldBuilder<T> = Widget Function(BuildContext context, T style, Set<Wi
...
+///
+/// [style] is the field's style.
+/// [states] is the current states of the widget.
+///
+/// The builder should use the provided `visibilityController` to toggle visibility
+/// and listen to visibility state changes.
+///
+/// See [FTextField.password].
+typedef FPasswordToggleBuilder = Widget Function(BuildContext context, FTextFieldStyle style, Set<WidgetState> states);
+
```
Pante: I think we can re-use `FFieldIconBuilder` instead of declaring a new builder.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -983,6 +920,194 @@ class FTextField extends StatefulWidget {
...
+  ///     icon: Icon(controller.value ? Icons.visibility_off : Icons.visibility),
+  ///   ),
+  /// )
+  ///
+  /// // Disable toggle completely
+  /// FTextField.password(
+  ///   toggleBuilder: null,
+  /// )
+  /// ```
+  static Widget password({
```
Pante: I think we can create a private subclass of FTextField. That way we can use a redirecting factory constructor thereby solving this issue.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -983,6 +920,194 @@ class FTextField extends StatefulWidget {
...
+    EditableTextContextMenuBuilder contextMenuBuilder = Defaults.contextMenuBuilder,
+    bool canRequestFocus = true,
+    UndoHistoryController? undoController,
+    SpellCheckConfiguration? spellCheckConfiguration,
+    FFieldIconBuilder<FTextFieldStyle>? prefixBuilder,
+    FFieldIconBuilder<FTextFieldStyle>? suffixBuilder,
+    bool Function(TextEditingValue) clearable = Defaults.clearable,
+    String? initialText,
+    FPasswordToggleBuilder? toggleBuilder = _defaultToggleBuilder,
+    FValueNotifier<bool>? visibilityController,
```
Pante: I think we can use a plain `ValueNotifier` here. Also I think we should rename this to `obscuredText`, kind of treating the notifier as a overload of the original `obscuredText`.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1187,359 @@ class FTextField extends StatefulWidget {
...
+    builder: (context, isVisible, child) {
+      // Determine the suffix builder based on priority:
+      // 1. Custom suffixBuilder (highest priority - disables toggle)
+      // 2. toggleBuilder (if not null, shows toggle using internal controller for default, requires external controller for custom)
+      // 3. null (no suffix)
+      final FFieldIconBuilder<FTextFieldStyle>? effectiveSuffixBuilder =
+          widget.suffixBuilder ??
+          (widget.toggleBuilder != null
+              ? (context, style, states) => _buildToggleWidget(context, style, states, isVisible)
+              : null);
```
Pante: I think we should either:
* Automatically wrap these in a row. 
* Change the API to accept either a suffix builder or toggle builder, not both.

Pante (commented):
Sorry for the late reply.

MoathCodes:
Hi,
- I removed the separate toggle-builder and now use the existing suffixBuilder; callers can disable the toggle by passing suffixBuilder: null.
- I implemented the password field as a private subclass.
~- I wanted the subclass to have a const constructor, so I avoided a redirecting factory with '=' and instead instantiate the private const subclass from the factory.~
~- One remaining issue: I couldn’t get parameter docs to appear on the factory when redirecting — I tried the redirecting '=' approach, but it didn’t solve the documentation visibility problem.~

What do you think?

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
   }
 }
 
+/// Internal password field implementation used by [FTextField.password].
```
Pante: I think we can move this into a separate file and mark it with `@internal`. The current text-field file's quite long.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
+///
+/// This widget wraps a regular [FTextField] and wires a visibility toggle.
+/// It exposes an optional [obscureTextController] ([ValueNotifier<bool>]) that
+/// when non-null lets external code observe and control whether the password
+/// is obscured. The controller's value mirrors [obscureText] directly
+/// (true = obscured/hidden).
+class _PasswordFieldWithToggle extends FTextField {
+  final ValueNotifier<bool>? obscureTextController;
```
Pante:
```suggestion
  final ValueNotifier<bool>? obscureText;
```

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
+      )
+      ..add(ObjectFlagProperty<EditableTextContextMenuBuilder>.has('contextMenuBuilder', contextMenuBuilder))
+      ..add(DiagnosticsProperty<bool>('canRequestFocus', canRequestFocus))
+      ..add(DiagnosticsProperty<UndoHistoryController?>('undoController', undoController))
+      ..add(DiagnosticsProperty<SpellCheckConfiguration?>('spellCheckConfiguration', spellCheckConfiguration))
+      ..add(ObjectFlagProperty<FFieldIconBuilder<FTextFieldStyle>?>.has('prefixBuilder', prefixBuilder))
+      ..add(ObjectFlagProperty<FFieldIconBuilder<FTextFieldStyle>?>.has('suffixBuilder', suffixBuilder))
+      ..add(ObjectFlagProperty<bool Function(TextEditingValue p1)>.has('clearable', clearable))
+      ..add(StringProperty('initialText', initialText))
+      ..add(DiagnosticsProperty<ValueNotifier<bool>?>('obscureTextController', obscureTextController));
```
Pante: We shouldn't need to redeclare most of this since they're already handled in `FTextField`.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
+        keyboardType: widget.keyboardType,
+        textInputAction: widget.textInputAction,
+        textCapitalization: widget.textCapitalization,
+        textAlign: widget.textAlign,
+        textAlignVertical: widget.textAlignVertical,
+        textDirection: widget.textDirection,
+        autofocus: widget.autofocus,
+        statesController: widget.statesController,
+        obscuringCharacter: widget.obscuringCharacter,
+        obscureText: isObscured, // Toggle based on controller (now mirrors obscureText)
```
Pante:
```suggestion
        obscureText: isObscured,
```

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
 }
 
+/// Internal password field implementation used by [FTextField.password].
+///
+/// This widget wraps a regular [FTextField] and wires a visibility toggle.
+/// It exposes an optional [obscureTextController] ([ValueNotifier<bool>]) that
+/// when non-null lets external code observe and control whether the password
+/// is obscured. The controller's value mirrors [obscureText] directly
+/// (true = obscured/hidden).
+class _PasswordFieldWithToggle extends FTextField {
```
Pante:
```suggestion
class _PasswordField extends FTextField {
```

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
+      ..add(ObjectFlagProperty<FFieldIconBuilder<FTextFieldStyle>?>.has('suffixBuilder', suffixBuilder))
+      ..add(ObjectFlagProperty<bool Function(TextEditingValue p1)>.has('clearable', clearable))
+      ..add(StringProperty('initialText', initialText))
+      ..add(DiagnosticsProperty<ValueNotifier<bool>?>('obscureTextController', obscureTextController));
+  }
+}
+
+class _PasswordFieldWithToggleState extends State<_PasswordFieldWithToggle> {
+  late ValueNotifier<bool> _obscureController;
+  bool _shouldDisposeController = false;
```
Pante: I think we can just track `widget.obscureText == null` instead of storing an additional field.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -1062,6 +1174,287 @@ class FTextField extends StatefulWidget {
...
+  }
+
+  Widget _buildToggleWidget(BuildContext context, FTextFieldStyle style, Set<WidgetState> states, bool isObscured) {
+    // When obscured == true show the "eye" icon (meaning "show password").
+    final icon = isObscured ? FIcons.eye : FIcons.eyeOff;
+
+    return FButton.icon(
+      onPress: () => _obscureController.value = !_obscureController.value,
+      style: FButtonStyle.ghost(),
+      child: Icon(icon, semanticLabel: isObscured ? 'Show password' : 'Hide password'),
```
Pante: We probably need to localize it. I think you can mark it with a todo for now and ill handle it once it's merged in.

Pante:
> I wanted the subclass to have a const constructor, so I avoided a redirecting factory with '=' and instead instantiate the private const subclass from the factory.
> One remaining issue: I couldn’t get parameter docs to appear on the factory when redirecting — I tried the redirecting '=' approach, but it didn’t solve the documentation visibility problem.

Could you elaborate on this? I don't quite follow.

MoathCodes:
> > I wanted the subclass to have a const constructor, so I avoided a redirecting factory with '=' and instead instantiate the private const subclass from the factory.
> > One remaining issue: I couldn’t get parameter docs to appear on the factory when redirecting — I tried the redirecting '=' approach, but it didn’t solve the documentation visibility problem.
> 
> Could you elaborate on this? I don't quite follow.

I'm sorry i was trying to do something with the redirecting of the constructor that i realized it won't work, now the factory method looks much cleaner.

I've also included a `password_field_test.dart` file that test the icon/semantics as well as the state of the whole toggling procedure.

forui/lib/src/widgets/text_field/password_field.dart
```diff
@@ -0,0 +1,253 @@
...
+/// is obscured. The controller's value mirrors [obscureText] directly
+/// (true = obscured/hidden).
+@internal
+class PasswordField extends FTextField {
+  final ValueNotifier<bool>? obscureTextNotifier;
+
+  PasswordField({
+    super.style,
+    FFieldBuilder<FTextFieldStyle>? builder,
+    super.label = const Text('Password'),
```
Pante, line 24:
```suggestion
    // TODO: Localize
    super.label = const Text('Password'),
```

forui/test/src/widgets/text_field/password_field_test.dart
```diff
@@ -0,0 +1,134 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:forui/forui.dart';
+import '../../test_scaffold.dart';
+
+void main() {
+  group('FTextField.password', () {
+    testWidgets('password toggle displays correct icons and semantic labels', (tester) async {
+      final obscure = ValueNotifier<bool>(true);
+      addTearDown(obscure.dispose);
```
Pante:
```suggestion
      final obscure = autoDispose(ValueNotifier<bool>(true));
```

forui/test/src/widgets/text_field/password_field_test.dart
```diff
@@ -0,0 +1,134 @@
...
+
+      // Back to initial state
+      expect(find.bySemanticsLabel('Show password'), findsOneWidget);
+      expect(find.byIcon(FIcons.eye), findsOneWidget);
+      expect(find.bySemanticsLabel('Hide password'), findsNothing);
+      expect(find.byIcon(FIcons.eyeOff), findsNothing);
+
+      final editableTextFinal = tester.widget<EditableText>(find.byType(EditableText));
+      expect(editableTextFinal.obscureText, isTrue);
+    });
```
Pante: I think this part should be separated into another test. What we really want to test here is that pressing the button when text is not obscured will obscure it.

forui/test/src/widgets/text_field/password_field_test.dart
```diff
@@ -0,0 +1,134 @@
...
+
+    testWidgets('multiple password fields share same obscure state', (tester) async {
+      final obscure = ValueNotifier<bool>(true);
+      addTearDown(obscure.dispose);
```
Pante:
```suggestion
      final obscure = autoDispose(ValueNotifier<bool>(true));
```

forui/test/src/widgets/text_field/password_field_test.dart
```diff
@@ -0,0 +1,134 @@
+import 'package:flutter/material.dart';
+
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:forui/forui.dart';
+import '../../test_scaffold.dart';
+
+void main() {
```
Pante, line 8: We should probably add a golden test for when the text is not obscured inside the text_field_golden_test file.

forui/lib/src/widgets/text_field/text_field.dart
```diff
@@ -983,6 +910,118 @@ class FTextField extends StatefulWidget {
...
+  ///   suffixBuilder: (context, style, states) => IconButton(
+  ///     onPressed: () => obscure.value = !obscure.value,
+  ///     icon: Icon(obscure.value ? Icons.visibility : Icons.visibility_off),
+  ///   ),
+  /// )
+  ///
+  /// // Disable the toggle completely
+  /// FTextField.password(suffixBuilder: null)
+  /// ```
+  factory FTextField.password({
```
Pante, line 959: Remember to expose it in `FTextFormField` as well!

forui/test/src/widgets/text_field/password_field_test.dart
Pante, line 8: nit: for assertions we generally prefer use `true` and `false` directly over `isTrue` and `isFalse`.

MoathCodes:
> Generally LGTM, just a few minor nits (Remember to update the README.md too).

I'm sorry which README.md file exactly?

Pante:
My bad, CHANGELOG.md not README.md*

MoathCodes:
It's my first time dealing with golden test hopefully I've done it right.

Another thing I see that the FTextFormField.password can be simplified if we use the FTextField.password but I couldn't find a way because i kept getting into problems with the `_FormField` class, I was afraid that I could break the form state I had a couple of ideas like passing an optional builder to the _FormField class that if not null will build FTextField.password otherwise it behaves as is, What do you think?

Pante:
Hmm in that case, I'll add the `FTextFormField.password` variant in a separate PR alongside the localization changes.

Pante (approved):
LGTM! Thanks for the PR!

MoathCodes:
Happy to help with this amazing project!

Pante:
I'm probably going to merge this PR later during the weekends. Just so I remember to do the remaining stuff.

MoathCodes:
I just realized that I've committed the edits to the sandbox file, I was testing and forgot to restore the file sorry.

PR #687, opened 2025-09-10
Make toaster api closer to scaffold messenger

Rexios80:
The current code requires direct access to a `BuildContext` to show a toast. This is unnecessary if you have access to an `FToasterState`. This PR makes the `FToaster` API behave more like the `ScaffoldMessenger` API.

Also I suggest committing your generated code so that builds are reproducible and so that you can use `forui` as a git dependency, but that is outside of the scope of this PR.

**Checklist**

### DIFF
--- a/forui/CHANGELOG.md
+++ b/forui/CHANGELOG.md
@@ -88,7 +88,7 @@ We've added password visibility toggle functionality to password fields.
 
 ### `FToaster`
 * Add `FToaster.of(...)`.
-* Make `context` optional in `FToasterState.show(...)`.
+* Make `FToasterState.show(context: ...)` optional.
 
 
 ## Others


Pante:
Thanks for the PR! 

That's odd but the internal gen tests are failing. It shouldn't be related to your changes, will have to look into it.
EDIT: Seems like a new dependency version changed the output's formatting.

> Also I suggest committing your generated code so that builds are reproducible and so that you can use forui as a git dependency, but that is outside of the scope of this PR.

We have a https://github.com/forus-labs/forui/tree/build-artifacts branch that automatically builds a version of main that can be used as a git dependency. That said, I think we could def improve visibility on it, maybe by renaming it to `nightly` or smth more obvious.

forui/CHANGELOG.md
```diff
@@ -86,6 +86,10 @@ We've added password visibility toggle functionality to password fields.
 * Add `FWidgetStateMap.lerpTextStyle(...)`.
 * Add `FWidgetStateMap.lerpWhere(...)`.
 
+### `FToaster`
+* Add `FToaster.of(...)`.
+* Make `context` optional in `FToasterState.show(...)`.
```
Pante:
```suggestion
* Make `FToasterState.show(context: ...)` optional.
```

Pante:
@Rexios80 Just a follow-up on this, it seems like I forgot to update the GHA to add the generated files after changing their extension. It should be fixed on the nightly branch once the PR I'm currently working on lands.

Issue #317, opened 2024-12-19
Tree


ZorinFoss:
A tree is a way of displaying a hierarchical list of items.

![tree](https://github.com/user-attachments/assets/4a422e42-6df2-457d-aad6-6610deb93288)

KrisBraun:
FSidebarItem achieves some of this, plus allows collapsing and expanding children.

A more general FTree would be useful, including capabilities like reordering items in the tree view with drag-and-drop and accessible keyboard shortcuts. @Pante suggested [some approaches to reordering here](https://github.com/forus-labs/forui/issues/702#issuecomment-3354489714).

I'm open contributing a PR once we have an approach we like.

Pante:
Copied from the prev issue:

## API/Implementation

The following will probably simplify the drag logic:
* https://api.flutter.dev/flutter/widgets/LongPressDraggable-class.html
* https://api.flutter.dev/flutter/widgets/Draggable-class.html
* https://api.flutter.dev/flutter/widgets/DragTarget-class.html

For the API, we can probably take inspiration from https://api.flutter.dev/flutter/material/ReorderableListView-class.html.

It will probably make sense to separate the logic from the visual appearance (except for the re-ordering animation). For the collapsing/expanding, we can probably build on top of `FCollapsible`. 

Shortcuts-wise, I've no idea how that interaction usually works with reorderables, open to feedback on this.

## Visual Design

I'm in favor of something like https://motion.dev/docs/react-reorder

PR #712, opened 2025-10-05
[Draft] FTree

KrisBraun:
**Describe the changes**

An initial attempt at #317. So far, it's simply a tree with line drawing (which was most of the challenge). I'll admit I relied heavily on Claude Code to bootstrap the new widget and haven't reviewed it all in detail yet.

Image

At this point, it would be helpful to get feedback on the widget API and design. I see a few open UI design questions:

1. Should the expand/collapse arrow be at the left or the right? On the left, it necessarily replaces any icon. On the right, it would either replace the icons shown in the original screenshot, or push them left.
2. How should expand/collapse interact with `onPress`? In my use case, I'd like clicking on the main `FTreeItem` to always expand and trigger `onPress` (since it's a selection of the current item), with a click on the icon required to collapse it. Perhaps if `onPress` is `null` then clicking anywhere always toggles.
3. How should the hover highlight interact with the lines? Since both are using the same color currently, the hover state makes the line disappear, which looks a bit off.

My intent is to create a separate PR to add reordering support after an initial FTree is merged. 

**Checklist**


KrisBraun:
I fully welcome contributions (or replacements if I've gone the wrong direction). Given my limited time to work on this, I'd be very happy for anyone to step in to make improvements directly. Of course, if no one else has interest or capacity, I'll get to feedback as soon as I can.

forui/test/golden/macos/tree/zinc-dark/nested.png
KrisBraun, line 1: I'm a bit stumped why this is showing an earlier version; when I put the test in the sample, it renders correctly. I've run `flutter test test/src/widgets/tree/tree_golden_test.dart --update-goldens` but it generates the same image.

Pante:
Both Joe & I will have limited capacity for most of October and early November. 

> Should the expand/collapse arrow be at the left or the right? On the left, it necessarily replaces any icon. On the right, it would either replace the icons shown in the original screenshot, or push them left.

I think the expand/collapse arrow should on the LHS for LTR locales. We should probably reverse this for RTL locales. IMO it shouldn't replace any user supplied icon. Rather than it should push the user supplied icon to the right.

> How should expand/collapse interact with onPress? In my use case, I'd like clicking on the main FTreeItem to always expand and trigger onPress (since it's a selection of the current item), with a click on the icon required to collapse it. Perhaps if onPress is null then clicking anywhere always toggles.

IMO we should definitely make this configurable via a parameter. Maybe something like an `FTreeItemExpandRegion` enum (enum name needs work) with icon & entire options? I think It make sense to default it to what you mentioned. 

> How should the hover highlight interact with the lines? Since both are using the same color currently, the hover state makes the line disappear, which looks a bit off.

Ideally the line should remain even when hovered. We probably need to play around with the colors to get the desired effect.

Pante:
The implementation has a couple of minor nits but looks okay in general! I think one important thing to note is that we need to add support for RTL locales (Basically mirror the current line drawing logic).

KrisBraun:
Thanks, I've added RTL support. I'll keep picking away at improvements.

KrisBraun:
I no longer need this in my project and haven't found the time to come back to it. I'm fine with this PR being closed for now, or finished up by someone else. (I'm very grateful for this project and will aim to contribute other ways in the future!)

Pante:
No worries! We’ll probably try to pick this up in due time.

Pante:
I'm closing this branch because it is really outdated and it'll be easier to copy these changes onto a new branch. This does not mean that we're not going to work on it.

PR #967, opened 2026-04-12
Adds generics support to FAutocomplete

chris-rutkowski:
Example usage:

```
class Fruit {
  final String name;
  final int calories;

  const Fruit(this.name, this.calories);

  @override
  String toString() => '$name calories $calories';
}

const fruits = [
  Fruit('Apple', 52),
  Fruit('Apple', 42),
  Fruit('Banana', 89),
  Fruit('Orange', 47),
  Fruit('Grape', 69),
  Fruit('Strawberry', 32),
  Fruit('Pineapple', 50),
];


          FAutocomplete<Fruit>.builder(
            hint: 'Type to search fruits',
            onSelect: (fruit) {
              print('Selected: ${fruit.name} with ${fruit.calories} calories');
            },
            filter: (query) async {
              await Future.delayed(const Duration(seconds: 3));
              return query.isEmpty
                  ? fruits
                  : fruits.where((fruit) => fruit.name.toLowerCase().startsWith(query.toLowerCase()));
            },
            displayStringForOption: (fruit) => fruit.name,
            contentBuilder: (context, query, values) => [
              for (final fruit in values)
                .item(value: fruit.name, subtitle: Text('${fruit.calories} calories'), data: fruit),
            ],
          )
```

Some info:

- no existing stuff is broken because if filter returns the array of strings - what was there by default, the generic type is implied to be String
- `displayStringForOption` default implementation returns `toString()`, which when casted on "legacy" implementation - only Strings would return those Strings as toString on string returns itself. If I didn't include this method in the snippet the placeholder string would display the name of fruit with calories, but obviously in placeholder i want just a name
- optional `onSelect` allows us to know what user clicked from the autocompletion list and is invoked also when user presses "tab"
- AI reviewer complained that the constructor of FAutocomplete or related object can no longer be constant constructor, but I don't see a single reasonable implementation of any form widget being constant.

Some screenshots:

Screenshot 2026-04-12 at 14 22 08

Screenshot 2026-04-12 at 14 22 10

Screenshot 2026-04-12 at 14 22 13

### DIFF
--- a/forui/CHANGELOG.md
+++ b/forui/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 0.21.0 (Next)
 
+### `FAutocomplete`
+* Add support for generic autocomplete options via `FAutocomplete<T>`.
+* Add `FAutocomplete.displayStringForOption` and `FAutocomplete.onSelect`.
+
+
 ### `FTypography`
 * Add `FTypography.defaultFontFamily` static constant.
 
--- a/forui/lib/src/foundation/typeahead_controller.dart
+++ b/forui/lib/src/foundation/typeahead_controller.dart
@@ -123,6 +123,7 @@ class FTypeaheadController<T> extends TextEditingController {
     text ??= this.text;
     if (text.isEmpty) {
       current = null;
+      currentData = null;
       return;
     }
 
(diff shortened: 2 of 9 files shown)


Pante:
I'm going to decline this PR. This is not because I don't see the use-case behind making `FAutocomplete` generic but rather while I don't have anything concrete at the moment, I believe it is worth re-investigating whether a generic `FAutocomplete` can be extended to supported more use-cases and whether it is better to split off into another widget like `FSearchableSelect`.

Issue #482, opened 2025-04-16
AnimatedTappableState._onPointerUp error


antons-zubkovs:
**Describe the bug**
I get following errors logged into Firebase Crashlytics:

```
          Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Null check operator used on a null value
       at AnimationController.stop(animation_controller.dart:924)
       at AnimationController._animateToInternal(animation_controller.dart:690)
       at AnimationController.reverse(animation_controller.dart:538)
       at AnimatedTappableState._onPointerUp(tappable.dart:329)
       at _FTappableState.build.<fn>(tappable.dart:210)
       at RenderPointerListener.handleEvent(proxy_box.dart:3097)
       at GestureBinding.dispatchEvent(binding.dart:499)
       at RendererBinding.dispatchEvent(binding.dart:460)
       at GestureBinding._handlePointerEventImmediately(binding.dart:437)
       at GestureBinding.handlePointerEvent(binding.dart:394)
       at GestureBinding._flushPointerEventQueue(binding.dart:341)
       at GestureBinding._handlePointerDataPacket(binding.dart:308)
```

My app uses FButton and FTappable for every clickable widgets, including those which trigger app navigation, which could be the cause of the issue (animation controller is disposed before `_onPointerUp()` is called).

**To Reproduce**
I was not able to reproduce it on my devices. 
There is no specific user journey which leads to the error; happens at different times/screens.
Could be artificially reproduced by adding delay in `_onPointerUp()` before accessing animation controller.

**Expected behavior**
Error is not thrown / `_onPointerUp()` has safety check before accessing animation controller.

**Device (please complete the following information):**
 - Flutter version: 3.29.2
 - Forui version: 0.10.0+1
 - Device (Android version): Galaxy Tab E 8.0 (7.1.1); Moto G 5G Plus (11); Redmi Note 8 (9); Nokia C12 (12); Galaxy M34 5G (14); Model:Galaxy A15 (14); Galaxy J7(2016) (8.1.0)

Pante:
This is an interesting issue, I'm unable to replicate it on my side as well without introducing an artificial delay in `_onPointerUp`. 

The only explanation I have for this is that there's some weird race condition between several different tappables or  Flutter is calling the `onPointerUp`  much later than it.

Either ways, I've added a `mounted` check before firing the controller.

PR #1063, opened 2026-06-18
Fix FTappable null check crash when unmounted mid-gesture in a group

elrizwiraswara:
**Describe the changes**

- Fix `FTappable` throwing `Null check operator used on a null value` when it is unmounted mid-gesture while inside an `FTappableGroup`.
- Root cause: `FTappableGroup`'s recognizer keeps a reference to a tappable's `GroupEntry` and invokes its `exit` callback (`_cancel`) on pointer move/cancel/reject. If the tappable was unmounted during the gesture (e.g. an item removed from a list while pressed), `_cancel` called `setState` on the defunct `State`, hitting `_element!.markNeedsBuild()`, which is the null check crash in release/profile mode.
- Fix: guard `_cancel` with a `mounted` check before touching `widget`/`setState`, matching the existing guards in `_start` and `_end` (and the issue #482 pattern).
- Added a regression test (`in FTappableGroup entry unmounted mid-gesture does not crash on cancel`) that reproduces the crash on the old code and passes with the fix.
- Updated `CHANGELOG.md` under a new `FTappable` & `FTappableGroup` section.

**Checklist**
- [x] I have read the CONTRIBUTING.md.
- [ ] I have included the relevant samples. (N/A, internal crash fix, no API/sample change)
- [ ] I have updated the documentation accordingly. (N/A, no API change)
- [ ] I have added the required flutters_hook for widget controllers. (N/A, no controller change)
- [x] I have updated the CHANGELOG.md if necessary.

### DIFF
--- a/forui/CHANGELOG.md
+++ b/forui/CHANGELOG.md
@@ -102,6 +102,10 @@ and an API similar to other widgets.
 * Fix `FSwitch` registering two focus nodes, causing it to be a duplicate stop when tabbing through.
 
 
+### `FTappable` & `FTappableGroup`
+* Fix `FTappable` throwing a null check error when unmounted mid-gesture inside an `FTappableGroup`.
+
+
 ### `FTypography` & `FTypeface`
 
 We've added support for multiple typefaces by introducing a `FTypeface` that contains a text scale and changing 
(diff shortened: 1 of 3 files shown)


Pante:
Thanks for the fix!

PR #1070, opened 2026-06-20
Add internal theme preset codec

Pante:
**Describe the changes**

- Add `lib/src/theme/codec.dart`, an internal codec (not part of the public API; imported by full path with the lint ignored) shared by the upcoming create app and CLI.
- Define the customization catalogs as `code`-bearing enums implementing a shared `Code` interface: `BaseColor`, `PrimaryColor`, `FontFamily` (+ `FontFamilyCategory`), `IconLibrary`, and `Radius`.
- Add a `Preset` model with `encode()` and a `Preset.decode(...)` factory, serializing a preset to/from a 6-char string of codes in the order: base color, primary color, display font, body font, icon library, radius.
- `Preset.decode` throws a `FormatException` on an invalid length or an unknown code.
- Add `test/src/theme/codec_test.dart` covering encode order, decode, round-trips, and the error cases.

**Checklist**

🤖 Generated with [Claude Code](https://claude.com/claude-code)


forui/lib/src/theme/codec.dart
kawaijoe, line 1: Nit, but should this be under create or theme? In the future we could add non theme related presets as well.

forui/lib/src/theme/codec.dart
```diff
@@ -0,0 +1,538 @@
...
+    required this.icon,
+    required this.radius,
+  });
+
+  factory Preset.decode(String encoded) {
+    if (encoded.length != 6) {
+      throw FormatException('A preset must be exactly 6 codes long.', encoded, encoded.length);
+    }
+
+    // This is not optimal since it is O(n), but realistically it doesn't matter as n is small.
```
kawaijoe: Technically its O(1) :p, since n is constant.

forui/lib/src/theme/codec.dart
```diff
@@ -0,0 +1,538 @@
...
+  final String url;
+
+  const IconLibrary(this.code, this.name, this.package, this.url);
+}
+
+enum Radius implements Code {
+  none(
+    'a',
+    'None',
+    FBorderRadius(xs2: .zero, xs: .zero, sm: .zero, md: .zero, lg: .zero, xl: .zero, xl2: .zero, xl3: .zero),
```
kawaijoe: Missing `pill` as border radius.

forui/lib/src/theme/codec.dart
Pante: I think this should be under theme, reason being that putting this in theme will introduce a cyclic dependency between forui and create. Not to mention, this will cause forui to transitively depend on hugeicons, iconoir etc.

forui/lib/src/theme/codec.dart
Pante: pill defaults to round, i dont think it should change with the various different border radii.

forui/lib/src/theme/codec.dart
kawaijoe: Oh what I meant was the folder, `src/theme` vs `src/create` or `src/util`.

forui/lib/src/theme/codec.dart
kawaijoe: Screenshot 2026-06-21 at 8 53 08 PM

Screenshot 2026-06-21 at 8 53 30 PM


I don't think that's right.

forui/lib/src/theme/codec.dart
Pante: no strong opinion, I'll move it under create.

forui/lib/src/create/codec.dart
```diff
@@ -0,0 +1,715 @@
...
+      errorForeground: 0xFFFAFAFA,
+      card: 0xFFFFFFFF,
+      border: 0xFFE8E4E3,
+    ),
+    dark: BaseColors(
+      barrier: 0x7A000000,
+      background: 0xFF0C0A09,
+      foreground: 0xFFFBFAF9,
+      primary: 0xFF1D1816,
+      primaryForeground: 0xFFFBFAF9,
```
kawaijoe: According to Claude, the primary colors here are incorrect. I've manually verified and the image below will render. The primary color should be a light color and the foreground should be a dark color in dark mode.

Image

kawaijoe (commented):
Generally looks good. One issue was flagged out when I got Claude to review the PR.

kawaijoe (approved):

PR #1162, opened 2026-08-18
Add `autocompleteEnabled` to `FAutocomplete`

### LINKED ISSUE #1034: Ability to turn of FAutocomplete at a runtime
While `FAutocomplete` is great I need an ability to disable in a runtime both:
- suggestion list popover
- typeahead (placeholder autocompleting the first item)

E.g. scenarios:
- user presses ignore suggestions button
- selection from autocompletion is allowed to be triggered only once (especially after this issue https://github.com/duobaseio/forui/issues/1033 where the autocompletion scope can be bigger)
- autocompletion enabled only in the form for "adding new item", but not for "editing existing one"

I'd like the `FAutocomplete` or `FAutocompleteController` to expose a boolean/method to disable these features and literally convert to ordinary text field. When disabled, if popover or typeahead are currently displayed, popover disappears, and so does typeahead.

Please note1: `FAutocomplete` contains property `enableSuggestions` but it is for something very different.

Please note2: I tried to cook some rudimentary solution for this, but it was glitching. Options I tried were:

- having my own `suggestionEnabled` prop, calling setState and returning empty list from `filter` and `contentBuilder`, but those two are not triggered when user inputs stays the same
- for above solution I tried to add a key, `FAutocomplete(key: ValueKey('autocomplete_$suggestionEnabled'))`, but it was causing exceptions saying that notifiers are used after being disposed in various scenarios.

GhagSagar23:
**Describe the changes**

- Closes #1034.
- Adds a backward-compatible `autocompleteEnabled` property, defaulting to `true`, to every `FAutocomplete` constructor.
- Addresses autocomplete state that previously remained loaded until text or focus changed, preventing a parent rebuild from immediately suppressing the popover and inline typeahead.
- When disabled, clears autocomplete suggestions, invalidates pending popover updates, and dismisses the popover without changing the field's text, selection, or focus.
- Prevents asynchronous results from reopening autocomplete while it is disabled and reloads autocomplete state when re-enabled.
- Adds widget coverage for initially disabled, enabled-to-disabled, and disabled-to-enabled behavior.
- Updates the API usage snippets and unreleased changelog.
- Leaves controller-level enablement and generalized filter invalidation outside this change.
- Risk: Flutter tests and analysis remain unverified because the configured Docker/Colima daemon was unavailable.

**Checklist**

- [x] A Flutter hook update is not applicable because no controller API was added.


Pante:
Thanks for opening a PR! Unfortunately, we're temporarily suspending merging of pull requests until migration to 3.47 has been completed.

Pante:
Thanks for this, and sorry for the slow turnaround.

I've landed `enableAutocompletion` in #1166 instead. You're credited as a co-author on the commit as several of the test
scenarios came from here.

The rewrite was mostly structural and I felt the feedback may help:

* The naming wasn't ideal and didn't follow existing conventions, e.g. `autocompleteEnabled` follows a different naming
  from `enableSuggestions`. It's even more jarring since they were placed 2 lines apart.
* The gating logic is scattered throughout the widget, this makes reasoning about the logic difficult.

